PR #34, opened 2015-07-30
update embargo release cron

rosy1280:
we only want the embargo release cron to run on one machine, not all of them.


rosy1280:
so it turns out this still isn't working, its only deploying to one server, the server with the db role.

PR #38, opened 2015-10-01
bump dor-services to 4.22.2

drh-stanford:


drh-stanford:
@rosy1280 please take a look at this...

PR #42, opened 2016-01-26
bundle update nokogiri

rosy1280:


jmartin-sul:
looks reasonable and tests pass.  mini_portile makes a big version jump, but in service of a minor nokogiri version bump, so that seeems fine.

PR #50, opened 2016-03-02
Use workflow client from Dor::Config

cbeer:


darrenleeweber:
Not a big fan of the indirection introduced, but there seems to be no stopping this train.

PR #53, opened 2016-03-03
Add basic accession robots class that provides the correct wiring for…

cbeer:
… the workflow service


drh-stanford:
In the current deploy, I'm doing this:
```

**Comment by cbeer** [comment]
:+1: that's one way to do it, yes. I've also proposed doing that globally again in https://github.com/sul-dlss/dor-services/pull/194.

I'm not sure it's a good long term strategy, though. It feels very magical and indirect.

File: robots/dissemination/cleanup.rb [hunk 1]
```diff
@@ -26,6 +26,10 @@ def is_full_druid_tree(druid)
           truncate_druid_tree = DruidTools::AccessDruid.new(druid, workspace_root)
           (!Dir.glob(truncate_druid_tree.path).empty? && !Dir.glob(full_druid_tree.path + '*').empty?)
         end
+
+        def workflow_service
```
drh-stanford: shouldn't we follow the same pattern and create a `Robots::DorRepo::Dissemination::Base` that has this method?

cbeer: This seems to be the only Dissemination robot. Do you know if we anticipate more?

drh-stanford: I dunno. So you prefer to defer until we have more than one?

cbeer: That'd be my slight preference. Happy to be persuaded, though.

drh-stanford: I like the clarity of `Robots::DorRepo::Accession::Base` and it's a little confusing as to why the cleanup class has the new code, whereas it makes more sense in a Base class. If we don't put it in a Base class, then maybe a comment as to why it's in the cleanup class (i.e., "move this into a base class if we get more dissemination robots").

cbeer: Ok. Let's split out a `Dissemination::Base`, then.

PR #59, opened 2016-06-07
Run cron job from within correct directory.

tingulfsen:


drh-stanford:
This correctly generates a crontab entry of:

```
16 2 * * * /bin/bash -l -c 'cd /home/lyberadmin/common-accessioning/current/ && BUNDLE_GEMFILE=/home/lyberadmin/common-accessioning/current/Gemfile ROBOT_ENVIRONMENT=production /usr/local/rvm/wrappers/default/ruby /home/lyberadmin/common-accessioning/current/robots/accession/embargo_release.rb >> /home/lyberadmin/common-accessioning/current/log/crondebug.log 2>&1'
```

PR #62, opened 2016-06-14
upgrade dor-services, prompt for branch to deploy

jmartin-sul:
* move to the current latest version of dor-services (common-accessioning robot configs are being updated to be in line with the latest dor-services version, by removing explicit configuration of the SDR URL, now that SDR access is handled via dor-services-app proxy).
* prompt for the branch to deploy, default to the branch of the current working directory.


ndushay:
in future, please include the version you upgraded to in the commit comment.   I look at commit history in github a lot and it's super helpful.

Issue #63, opened 2016-06-15
embargo_release not resulting in correct rightsMD


ndushay:
proxied by sul-dlss/media#80

embargo-release is failing to either:

- properly copy the new rightsMD from embargoMD at time of release -OR-
- properly save the updated rightsMD datastream in the versioning process

Ben has successfully released objects from embargo using the following:

1. run script below [1]
2. prepare new version for the objects in Argo
3. close version for the objects in Argo (note - I have stripped out the open_version and close_version methods commented out in the script below)

This has successfully performed the release as expected.

[1]
```ruby
embargo_druids.each do | druid |
  item=Dor::Item.find("druid:#{druid}")
  #item.open_new_version
  item.release_embargo("application::accessionWF:embargo-release")
  #item.close_version :description => "embargo released", :significance => :admin
  item.save
end
```

possibly more info:
sul-dlss/media#70 
sul-dlss/media#68 
#60

ndushay:
An embargoed datastream looks like this
```
<embargoMetadata>
  <status>embargoed</status>
  <releaseDate>2016-06-13T08:00:00Z</releaseDate>
  <twentyPctVisibilityStatus/>
  <twentyPctVisibilityReleaseDate/>
  <releaseAccess>
    <access type="discover">
      <machine>
        <world/>
      </machine>
    </access>
    <access type="read">
      <machine>
        <world/>
      </machine>
    </access>
  </releaseAccess>
</embargoMetadata>
```

Embargo release replaces the object level rights in rightsMetadata with the contents of \<releaseAccess> and then sets the status to "release". 

rightsMetadata for an item under embargo looks like this: 
```
<rightsMetadata>
  <access type="discover">
    <machine>
      <world/>
    </machine>
  </access>
  <access type="read">
    <machine>
      <embargoReleaseDate>2016-06-13T08:00:00Z</embargoReleaseDate>
      <none/>
    </machine>
  </access>
  ...etc...
```

The two expressions of release date should match of course, and be in the past to be caught by the embargo release job's query.

The rightsMD SHOULD look like this after embargo is released by the script:
```
<rightsMetadata>
  <access type="discover">
    <machine>
      <world/>
    </machine>
  </access>
  <access type="read">
    <machine>
      <world/>
    </machine>
  </access>
...
```

ndushay:
[...]

PR #64, opened 2016-06-16
embargo_release: make it a class and write tests

ndushay:
(sadly, this did NOT allow us to get the failures with release_embargo we are experiencing in situ, but it still seems like a grand idea to have specs.)

@peetucket @atz   might you review this?  Thanks. @cbeer @drh-stanford ?  Anyone?  it's pretty simple.

connects to #63

### DIFF
--- a/spec/robots/accession/embargo_release_spec.rb
+++ b/spec/robots/accession/embargo_release_spec.rb
@@ -1,0 +1,132 @@
+require 'spec_helper'
+
+describe 'embargo_release.rb' do
+  Dor.configure do
+    workflow.url 'http://example.org/workflow'
+  end
+
+  before(:each) do
+    allow_any_instance_of(RSolr::Client).to receive(:find) {
+      {
+        'response' => {
+          'numFound' => '1',
+          'docs' => {}
+        }
+      }
+    }
+    # must do require after mocking Solr call because loading the file calls method
+    require File.expand_path(File.dirname(__FILE__) + '/../../../robots/accession/embargo_release')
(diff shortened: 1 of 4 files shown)


ndushay:
@jmartin-sul   I already reverted your temporary commit that allowed us to test this on robot-test.   This is ready to be merged.

spec/robots/accession/embargo_release_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+
+    it 'embargo metadata changes to twenty_pct_status released after Dor::Item.release_20_pct_vis_embargo' do
+      expect(item.embargoMetadata.twenty_pct_status).to eql 'anything'
+      item.release_20_pct_vis_embargo('ignored')
+      expect(item.embargoMetadata.twenty_pct_status).to eql 'released'
+    end
+  end
+
+  it 'running embargo_release.rb attempts to run EmbargoRelease.release' do
+
```
atz: skip or TODO?

jmartin-sul: wontfix, will remove (couldn't come up with a way to test that didn't require more refactoring, didn't seem worth expending tons of effort on since it'll be pretty obvious from trying to run the script whether it's doing something or nothing)

jmartin-sul: actually, made it a skip with an explanation and refined the test description.  otherwise, everything's the same as it was before.

ready for another look from @atz...

PR #76, opened 2016-10-24
switch from Dor::Item.find to Dor.find when object type is unknown a priori

### LINKED ISSUE #75: change calls to Dor::Item.find to Dor.find
We need to use `Dor.find` since we're not sure what the object type is during many (all?) steps of the accessioning workflow.

drh-stanford:
This PR fixes #75 and is in preparation to use ActiveFedora 7 which will throw exceptions on type mismatches (i.e., using `Dor::Item.find` for a `Dor::Collection` object).


cbeer:
👍

PR #83, opened 2017-04-20
do not call methods on objects if those methods are not defined

### LINKED ISSUE #82: Each robot should confirm that the method is defined before calling it
The updated version of ActiveFedora does not define methods for objects that don't need them (e.g. APO does not define "publish_metadata" or "shelve").  However, the robots will blindly call those methods on the objects after issuing a Dor.find.  For each robot that blindly calls a method, we should confirm the existence of that method before making the call.

Also true of calls to upgrade! for remediate-object (which is no longer defined in dor-services >= 5.17)

peetucket:
fixes #82


spec/robots/accession/publish_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  let(:druid) { 'druid:oo000oo0001' }
+
+  it 'includes behavior from LyberCore::Robot' do
+    robot = Robots::DorRepo::Accession::Shelve.new
+    expect(robot.methods).to include(:work)
+  end
+
+  it 'calls .publish_metadata if that method is defined (e.g. item)' do
+    object = double(:publish_metadata => true)
+    expect(object.respond_to?(:publish_metadata)).to be true
```
jmartin-sul: i think i'd be inclined to omit this, since it seems more like it's testing `double` than our method?

spec/robots/accession/publish_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    expect(Dor).to receive(:find).with(druid).and_return(object)
+    robot = Robots::DorRepo::Accession::Publish.new
+    expect(object).to receive(:publish_metadata)
+    robot.perform(druid)
+  end
+
+  it 'does not call .publish_metadata if that method is not defined (e.g. apo)' do
+    object = Object.new
+    expect(object.respond_to?(:publish_metadata)).to be false
```
jmartin-sul: ditto on testing `double`

spec/robots/accession/publish_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    expect(Dor).to receive(:find).with(druid).and_return(object)
+    robot = Robots::DorRepo::Accession::Publish.new
```
jmartin-sul: but, i would be inclined to put in something like `expect(object).not_to receive(:publish_metadata)`, kind of parallel to what's in the above test.

spec/robots/accession/shelve_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  let(:druid) { 'druid:oo000oo0001' }
+
+  it 'includes behavior from LyberCore::Robot' do
+    robot = Robots::DorRepo::Accession::Shelve.new
+    expect(robot.methods).to include(:work)
+  end
+
+  it 'calls .shelve if that method is defined (e.g. item)' do
+    object = double(:shelve => true)
+    expect(object.respond_to?(:shelve)).to be true
```
jmartin-sul: ditto on testing `double`

spec/robots/accession/shelve_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    expect(Dor).to receive(:find).with(druid).and_return(object)
+    robot = Robots::DorRepo::Accession::Shelve.new
+    expect(object).to receive(:shelve)
+    robot.perform(druid)
+  end
+
+  it 'does not call .shelve if that method is not defined (e.g. apo)' do
+    object = Object.new
+    expect(object.respond_to?(:shelve)).to be false
```
jmartin-sul: ditto on testing `double`

spec/robots/accession/shelve_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    expect(Dor).to receive(:find).with(druid).and_return(object)
+    robot = Robots::DorRepo::Accession::Shelve.new
```
jmartin-sul: and then here, i'd put `expect(object).not_to receive(:shelve)`

spec/robots/accession/publish_spec.rb
peetucket: Joe and I realized this won't work, since expect apparently defines the mock method you are testing (:publish_metadata) to confirm it won't get called...but then our logic thinks it exists, and calls it, causing the test to fail... it took us a while to sort this out, which means you can't test the expectation like that and instead needs to test via a non-exception

spec/robots/accession/shelve_spec.rb
peetucket: see comment above for publish_spec

spec/robots/accession/publish_spec.rb
peetucket: yeah, this was helpful for debugging, but realize its not an actual test, removing all of them

robots/accession/remediate_object.rb
```diff
@@ -11,7 +11,7 @@ def initialize
 
         def perform(druid)
           obj = Dor.find(druid)
-          obj.upgrade!
+          obj.upgrade! if obj.respond_to? :upgrade!
```
jmartin-sul: i *think* it's actually ok to just remove this call entirely, and drop the new tests, as i think the `upgrade!` capability was removed from dor-services late last year:
https://github.com/sul-dlss/dor-services/pull/260

peetucket: yes, indeed that functionality was removed...but there was some discussion with Lynn that it in the short to medium term was better to continue making the call if it exists just in case we wanted to re-implement some part of "upgrade!" later (robot becomes a noop now, but may resume doing something later)

if that doesn't happen in some intermediate term, we would remove the robot entirely

PR #102, opened 2018-04-18
dor-services v5.29.0 to use new preservationIngestWF

ndushay:
by using dor-services v5.29.0, common-accessioning robot `sdr-ingest-transfer` will kick off preservationIngestWF instead of sdrIngestWF, thus using the new code preservation_robots instead of sdr-preservation-core.

I assert that there is functional parity;  this has been tested by manually running a test object through common accessioning and preservation robots on production.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
-gem 'activesupport'
+gem 'activesupport', '~> 5.1.0' # 5.2.0 breaks: "can't modify frozen ActiveSupport::HashWithIndifferentAccess"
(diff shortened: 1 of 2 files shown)


peetucket:
Looks like maybe the bump to ActiveSupport is causing some issues with HashWithIndifferentAccess in the tests?  Did a quick google search and didn't turn up anything with Rails 5

ndushay:
@peetucket ja - looking into it.

jcoyne:
Ref https://github.com/samvera/active_fedora/issues/1307

PR #104, opened 2018-04-26
Directed at production problem related to dependency shift

atz:
ActiveFedora (in our old-ass version) still expects to be able to `return false` out of a callback to skip remaining callbacks.

This buys us some time to still use rails 5.1 with ActiveFedora, but it is fleeting, since the option goes away in rails 5.2.

See: https://github.com/sul-dlss/hydrus/issues/161

### DIFF
--- a/config/boot.rb
+++ b/config/boot.rb
@@ -43,3 +43,8 @@ begin
   end
 end
 require 'robot-controller'
+
+# ActiveFedora::OmDatastream v8.4.2 uses old-style `return false` to break callback cycle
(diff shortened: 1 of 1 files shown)


atz:
After deployment, restart the failed step in this object to test whether resolution is achieved:
https://argo.stanford.edu/view/druid:hp518bp9831

atz:
This seems to have been ineffective because in our version of activesupport, the option is deprecated (and nullified).

PR #106, opened 2018-06-04
Add build for ruby 2.3.7

tallenaz:
prep work for ruby upgrade in sul-robots-test


jcoyne:
Looks like this needs a rebase.

PR #123, opened 2018-11-30
Remove the integration_tests

jcoyne:
These tests are never run in CI because the have dependencies on external servers.
We don't have an environment where we can run these.


mjgiarlo:
Your thoughts on this, @cbeer 	@peetucket @SaravShah @atz 	?

jcoyne:
I'll also note that this make upgrading dor-services easier because we can drop the dependency on `assembly-utils`

SaravShah:
i'll let peter, chris, or joe respond. I've only touched this code base from updating dependencies :)

peetucket:
Don’t have a strong opinion either way. Coverage is pretty bad according to coveralls (and this doesn’t change it).

cbeer:
🤷‍♂️

PR #126, opened 2018-12-04
Move the shelving service from dor-services to common-accessioning

jcoyne:


robots/accession/shelve.rb
```diff
@@ -9,11 +12,9 @@ def initialize
 
         def perform(druid)
           obj = Dor.find(druid)
-          obj.shelve if obj.respond_to? :shelve
+          Dor::Shelve.push(obj) if obj.is_a? Dor::Item
```
mjgiarlo: Can you say more about this change? It's not clear as a reviewer that they are functionally equivalent.

jcoyne: The `shelve` method is deprecated on dor-services because it violates the SRP. https://github.com/sul-dlss/dor-services/pull/393

jcoyne: Dor::Item is the only object that includes `Shelvable`: https://github.com/sul-dlss/dor-services/blob/523b84df36376cc2a50cc08b45e26616ec8cc670/lib/dor/models/item.rb#L5

lib/dor/shelve.rb
```diff
@@ -0,0 +1,68 @@
...
+      stacks_object_pathname = stacks_location
+      # determine the location of the object's files in the stacks area
+      stacks_druid = DruidTools::StacksDruid.new work.id, stacks_object_pathname
+      stacks_object_pathname = Pathname(stacks_druid.path)
+      # determine the location of the object's content files in the workspace area
+      workspace_druid = DruidTools::Druid.new(work.id, Config.stacks.local_workspace_root)
+      workspace_content_pathname = workspace_content_dir(diff, workspace_druid)
+      # delete, rename, or copy files to the stacks area
+      DigitalStacksService.remove_from_stacks(stacks_object_pathname, diff)
+      DigitalStacksService.rename_in_stacks(stacks_object_pathname, diff)
```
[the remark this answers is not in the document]
jcoyne: This was code copied straight out of dor-services without changes.   So I'm just preserving the weirdness for this PR at least.

This is also why I needed to mess with rubocop, because this code was using Ruby 2.5 stuff, but the common-accessioning's rubocop was set to use Ruby 2.2 parser.  I tried to keep the changes separated into different commits. I think that's the best I can do without copying code in and changing it at the same time.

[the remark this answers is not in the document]
justinlittman: Ahh, missed the separate commits.

cmh2166:
Design question: why is this being moved?

jcoyne:
@cmh2166 and I discussed that an alternative approach, where the robot should make a call to dor-services-app to do the shelving.  This would make migration less difficult and allow the code to stay in dor-services.

PR #136, opened 2018-12-10
[WIP] Update to robot-controller 3.0.0.beta1

jcoyne:


jcoyne:
Robot-controller is no longer needed now that we use resque-pool

PR #158, opened 2019-01-16
Remove lite objects

jcoyne:
These are untested and I don't think they're used.


blalbrit:
This option was used to smoke test various -stage environment changes. Useful to be able to generate a bunch of objects for dev/testing purposes maybe. Probably haven't been used since 2015?

My recommendation is that we continue to move toward testing with real objects in the stage environment, rather than the lite objects, but I defer to @peetucket or other developers who have had experience with the lite objects.

PR #174, opened 2019-02-01
configure workflow service

peetucket:


config/boot.rb
```diff
@@ -33,6 +33,8 @@
 puts "Loading config from #{env_file}"
 require env_file
 
+Dor::WorkflowService.configure(Dor::Config::workflow.url)
```
jcoyne: Should this be in lyber-core?

peetucket: like this? https://github.com/sul-dlss/lyber-core/pull/35

peetucket: i guess we would presume that any consumer of the lyber-core gem can reasonably be expected to have a configuration that includes that URL

mjgiarlo:
Moved to lyber-core.

PR #186, opened 2019-02-08
Avoid the deprecated Item#shelve method

jcoyne:


peetucket:
looks good, one small rubocop complaint though

PR #215, opened 2019-04-11
Move ResetWorkspace from dor-services

jcoyne:


justinlittman:
Approve pending resolving conflicts.

PR #221, opened 2019-04-12
Move PublicXmlService from dor-services

jcoyne:


lib/public_xml_service.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  def initialize(object)
+    @object = object
+  end
+
+  def to_xml
+    pub = Nokogiri::XML('<publicObject/>').root
+    pub['id'] = object.pid
+    pub['published'] = Time.now.utc.xmlschema
+    pub['publishVersion'] = 'dor-services/' + Dor::VERSION
```
mjgiarlo: :eyes: do we care about this lying or meh?

mjgiarlo: Feel free to merge if meh.

jcoyne: I honestly have no clue as to what the consumer expects, so I don't want to change anything.

PR #235, opened 2019-04-17
Move provenance datastream building from dor-services

jcoyne:


robots/accession/provenance_metadata.rb
```diff
@@ -11,7 +11,34 @@ def initialize
 
         def perform(druid)
           obj = Dor.find(druid)
-          obj.build_provenanceMetadata_datastream('accessionWF', 'DOR Common Accessioning completed')
+          build_datastream(obj, 'accessionWF', 'DOR Common Accessioning completed')
+        end
+
+        private
+
+        # as well as Argo's ItemsController#fix_missing_provenance
```
mjgiarlo: I'm assuming you copied this over from dor-services and renamed the method from `#build_provenanceMetadata_datastream`? Does this comment mean moving this method out of dor-services will break argo? What's your plan for dealing with this?

mjgiarlo: Should this comment be deleted in this PR since Argo presumably won't be using a method defined in common-accessioning?

jcoyne: Yes, I'll delete the comment.  See https://github.com/sul-dlss/argo/pull/1366

PR #306, opened 2019-05-15
etd robots expect etd models

jcoyne:
Fixes https://github.com/sul-dlss/hydra_etd/issues/211


jcoyne:
@justinlittman better now?

PR #307, opened 2019-05-15
Fix run_marc_builder

tallenaz:


bin/run_marc_builder.sh
```diff
@@ -7,14 +7,4 @@
...
-# Make sure the kerberos ticket and afs token are renewed
-#
-echo "renew-ticket for afs"
-date
-KRB5CCNAME=/tmp/kerb5.tkt KINIT_PROG=/usr/bin/aklog \
-/usr/local/bin/k5start  -t -k /tmp/kerb5.tkt -H 60 -U -f \
-/opt/app/lyberadmin/sulair-lyberservices
-
-/opt/app/lyberadmin/common-accessioning/current/robots/etd_submit/build_symphony_marc.rb
+/opt/app/lyberadmin/common-accessioning/current/robots/dor_repo/etd_submit/build_symphony_marc.rb
```
jcoyne: Can we do away with run_marc_builder.sh and update the cron to just call that script ☝️

tallenaz: Yes, removed and updated.

PR #316, opened 2019-05-17
Reinstate run_marc_builder

tallenaz:
This time, in order to make marc deposits
  owned by the sirsi user

### DIFF
--- a/bin/run_marc_builder.sh
+++ b/bin/run_marc_builder.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+/opt/app/lyberadmin/common-accessioning/current/robots/dor_repo/etd_submit/build_symphony_marc.rb
+/bin/chown 214:214 /etd/*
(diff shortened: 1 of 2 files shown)


bin/run_marc_builder.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+/opt/app/lyberadmin/common-accessioning/current/robots/dor_repo/etd_submit/build_symphony_marc.rb
+/bin/chown 214:214 /etd/*
```
jcoyne: If these cron runs as lyberadmin, will it have permission to chown to another user?

tallenaz: That's right, the cron won't, and the FIleUtils approach throws and Operation not permitted error. An approach that could work would be to have a root user cron set in puppet to chown the file. In that case, we wouldn't need to reinstate the script. Closing this for now.

Issue #333, opened 2019-05-29
dissemination cleanup should use the dor-service-app endpoint


jcoyne:
See https://github.com/sul-dlss/dor-services-app/pull/300

Also remove CleanupResetService as it moved to dor-services-app
Then we can move Dor::CleanupService from dor-services to dor-services-app

peetucket:
This appears to be done?  The dissemination cleanup step in common-accessioning is hitting a `dor-services-client` defined endpoint.  See https://github.com/sul-dlss/common-accessioning/blob/master/lib/robots/dor_repo/dissemination/cleanup.rb

peetucket:
I also see a CleanupService in dor-services-app so perhaps that part is done too? https://github.com/sul-dlss/dor-services-app/blob/master/app/services/cleanup_service.rb

jcoyne:
Yes, this appears to be complete.

Issue #321, opened 2019-05-20
/dor mount not reliably cleared out by accessionWF reset-workspace


tallenaz:
Is this a dup of #342 ?  Is the fix similar?  Can we fix #342 at the same time, please?

<hr />

While debugging https://jirasul.stanford.edu/jira/browse/SDRO-266 @blalbrit and I discovered that the dor assembly directory is not cleaned out by reset-workspace as reliably as we expect.

blalbrit:
The directories that should be cleaned out will be:
* /dor/export
* /dor/workspace
* /dor/assembly

jmartin-sul:
might errors like this be related?
https://app.honeybadger.io/projects/52894/faults/49769895

`Errno::ENOENT: No such file or directory @ rb_file_s_rename - (/dor/workspace/zy/501/np/2944/zy501np2944, /dor/workspace/zy/501/np/2944/zy501np2944_v1)`
```
lib/reset_workspace_service.rb:11:in `reset_workspace_druid_tree`
lib/robots/dor_repo/accession/reset_workspace.rb:18:in `perform`
```

or is that a different problem?

if the latter, i can file a separate issue for figuring out whether there's a bug, or just a noisy alert.

jmartin-sul:
same question about e.g.:
https://app.honeybadger.io/projects/52894/faults/50132373

`Errno::ENOENT: No such file or directory @ dir_initialize - /dor/workspace/zy/501/np/2944`

```
lib/cleanup_reset_service.rb:39:in `cleanup_reset_workspace_content`
lib/cleanup_reset_service.rb:11:in `cleanup_by_reset_druid`
lib/robots/dor_repo/dissemination/cleanup.rb:14:in `perform`
```

peetucket:
Seems to be ongoing (new files with datestamps of May - July 2019)

peetucket:
Impact: If the /dor mount fills up over time due to the /dor/export and /dor/workspace not being cleared after accessioning/ingest is complete, a nagios alert goes out.  This requires some manual intervention to investigate and remove content and possibly some scripting to be sure we don't inadvertently remove content still in accessioning or versioning.  So the impact is mostly operations and developer time spent periodically tending to the drive filling up when it alerts.

There is no user facing impact, other than the possible pausing of accessioning if the /dor mount were to be filled up and not cleared manually.

ndushay:
[...]

PR #338, opened 2019-06-03
Use the workspace cleanup API in dor-services-app

jcoyne:
Connects to #333


ndushay:
Is this likely to help with dor mount cleanup issues (#321, #342) ?

And can you say something re: why it makes more sense for this workspace cleanup code to be in dor-services-app?

jcoyne:
This was part of the infrastructure plans to provide APIs for our business processes.  dor-services-app provides an API and this PR uses that API.  If we use APIs then we do not need to have shared code.  I do not think this has any bearing on #321 or #341.

ndushay:
thanks for the explanation!

PR #366, opened 2019-08-12
allow 3d content type to pass through from stub content metadata

peetucket:
ported from https://github.com/sul-dlss-deprecated/assembly/pull/66

This is a pre-req for the actual content metadata generation which occurs here: https://github.com/sul-dlss/assembly-objectfile/pull/7

### DIFF
--- a/spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
+++ b/spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Dor::Assembly::StubContentMetadataParser do
+  describe '#stub_content_metadata_parser' do
+    let(:druid) { 'druid:aa111bb3333' }
+    let(:item) { Dor::Assembly::Item.new(druid: druid) }
+
+    it 'maps content metadata types to the gem correctly' do
+      ['flipbook (r-l)', 'book', 'a book (l-r)'].each do |content_type|
+        allow(item).to receive(:stub_object_type).and_return(content_type)
+        expect(item.gem_content_metadata_style).to eq(:simple_book)
+      end
+      allow(item).to receive(:stub_object_type).and_return('image')
+      expect(item.gem_content_metadata_style).to eq(:simple_image)
+      allow(item).to receive(:stub_object_type).and_return('maps')
+      expect(item.gem_content_metadata_style).to eq(:map)
+      %w[3d 3D].each do |content_type|
+        allow(item).to receive(:stub_object_type).and_return(content_type)
+        expect(item.gem_content_metadata_style).to eq(:"3d")
+      end
+      %w[file bogus].each do |content_type|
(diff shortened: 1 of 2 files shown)


spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+require 'spec_helper'
+
+RSpec.describe Dor::Assembly::StubContentMetadataParser do
+  describe '#stub_content_metadata_parser' do
+    let(:druid) { 'druid:aa111bb3333' }
+    let(:item) { Dor::Assembly::Item.new(druid: druid) }
+
+    it 'maps content metadata types to the gem correctly' do
+      ['flipbook (r-l)', 'book', 'a book (l-r)'].each do |content_type|
```
jcoyne: I think these should be broken out into separate context blocks.  "When stub_object_type is book", etc.

spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+        allow(item).to receive(:stub_object_type).and_return(content_type)
+        expect(item.gem_content_metadata_style).to eq(:simple_book)
+      end
+      allow(item).to receive(:stub_object_type).and_return('image')
```
jcoyne: as well as this

spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      expect(item.gem_content_metadata_style).to eq(:simple_image)
+      allow(item).to receive(:stub_object_type).and_return('maps')
```
jcoyne: and this.

spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      expect(item.gem_content_metadata_style).to eq(:map)
+      %w[3d 3D].each do |content_type|
+        allow(item).to receive(:stub_object_type).and_return(content_type)
```
jcoyne: and these.

spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+        expect(item.gem_content_metadata_style).to eq(:"3d")
+      end
+      %w[file bogus].each do |content_type|
+        allow(item).to receive(:stub_object_type).and_return(content_type)
```
jcoyne: and these too of course 😄

peetucket:
Split spec into contexts

spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
peetucket: split them all out

PR #391, opened 2019-09-09
Notify the repo manager when an item can't be released

jcoyne:


config/settings.yml
```diff
@@ -1,4 +1,14 @@
...
-    url:  'https://dor-services-test.stanford.test'
-    token: secret-token
-    token_header: 'X-Auth'
+dor_services:
+  url:  'https://dor-services-test.stanford.test'
+  token: secret-token
+  token_header: 'X-Auth'
+
+repository_manager:
+  email: bergeraj@stanford.edu
```
ndushay: I doubt @andrewjbtw  wants this in a public repo.  Maybe make part of shared_configs?   Also, it would be better to get an indirect mailing list set up so it goes with the job, not the person.

andrewjbtw:
Justin and I did discuss this. It doesn't look like anything other than honeybadger is notifying if an embargo release fails. There may be an Argo error somewhere. I'd like a more deliberate notification, since it's easy to miss.

config/settings.yml
andrewjbtw: Yeah, I'd prefer not to have email address in the public code if possible.

jcoyne: @ndushay I don't know how to do that. Do you?

ndushay: i think you just add the value to https://github.com/sul-dlss/shared_configs/blob/common-accessioning-prod/config/settings/production.yml.    You can test it locally by adding it to config/settings/development.yml .

jcoyne: @ndushay I mean how do you "get an indirect mailing list set up" how does one get a list made?

ndushay: I think maybe one can be requested at lists.stanford.edu?

ndushay:
This is 2 years old - can it be closed and branch deleted?  (please do so)

config/settings.yml
ndushay: we could use dlss-infrastructure

ndushay: or @andrewjbtw is there an email for the sdr manager as a list?

jcoyne:
Embargo release is no longer in this repo.

andrewjbtw:
One of those tickets that came up when I'd worked here for like three weeks. Potentially still useful but we handle releases differently now and I look at HB a lot more.

PR #392, opened 2019-09-13
Remove old token header from settings.yml

jcoyne:


ndushay:
@jcoyne @mjgiarlo will this cause us the same problems we had in ... some other code base earlier this week ... re: token header stuff?  Should this PR be reverted before deployment monday?  Are there any others?

mjgiarlo:
I don't think so? I believe the problematic interaction was between the workflow server and dor-services-app.

Issue #386, opened 2019-09-06
error for "descMetadata missing required fields (<title>)" sometimes occurs when a title is present in descMetadata


andrewjbtw:
Some objects are stopping at descriptive-metadata with the following error

```descriptive-metadata : descMetadata missing required fields (<title>)```

but when users investigate, they find a descMetadata stream exists and has a title. Setting the step to "waiting" clears up the problem because the title is present at that point and the subsequent check is successful.

Approximately 50 ETDs have been caught at this step in the past two days. Some other users have seen the problem as well, outside of the ETD workflow.

It's possible that the check is running before the descMetadata datastream has been created.

I've put a question to non-ETD users to try to find out if this is always happening now or only sometimes happening but I have not heard back beyond the initial reports.

ndushay:
I looked over the code, and I checked the code with a colleague, and I think we *might* be at this point:

- Ben's uber ticket #188  said that sometimes retrieving stuff from Symphony came back with no results ... but was hard to track.   We put in this check so accessioning would fail on the correct step - the one where re-running the step can fix the problem.  

Earlier today, I found this error on a couple of objects in Argo (prod).  There was no descriptiveMetadata in the object;  I reset this step to waiting, and voila, accessioning happened.  However, the error didn't clear from the item view immediately -- I had to hit "reindex" -- which will happen on its own, eventually on some schedule I don't know.

I would be delighted to have an example of a specific case where this error appears AND reindexing doesn't clear the error AND the descriptiveMetadata is there and has a title.

andrewjbtw:
I was hoping for another example with all the conditions you describe, but nothing seems to have come through like that today.

However, the number of items in this state with no catkey and no descMetadata seems unusually high and makes me wonder if there's something else going on with how that datastream is created that might be unrelated to this ticket. I've tried resetting a number of these (no catkey, no descmetadata) and they get caught again. I've reindexed but not seen a change. Maybe it will look different Monday.

andrewjbtw:
I checked with the content owner on items that are missing the descMetadata datastream altogether and determined that is unrelated to this ticket. Certain old objects have this problem, so it was expected that those items would be caught at this step.

I will continue to keep an eye out for more objects that fail at this step and do have a title.

ndushay:
talk to @andrewjbtw for latest info;  we think the ETD workflow may trigger this.   https://consul.stanford.edu/display/ETD/ETD+Submission+Workflow

[...]

PR #394, opened 2019-09-14
FIX BUG:  desc metadata required fields check

### LINKED ISSUE #393: registered object with catkey and no label gets "no title" err from descMetadata step
Here's the error:

image

But descMetadata exists:

image

And has title:

image

So the fetch from Symphony worked ... but the check seems to have run first

related to #386

ndushay:
Fixes #393 
related to #386, and may fix that as well

We have been getting this error but inappropriately -- the descMetadata is there and it has a title:

image

I was able to reproduce the error in stage consistently by registering an object with a catkey and no label, and then adding the accessioningWF.

This PR fixes the problem by reloading the Dor object before checking for required fields in a possibly newly updated datastream.

I deployed the fix to stage and empirically established that the fix works.

### DIFF
--- a/lib/robots/dor_repo/accession/descriptive_metadata.rb
+++ b/lib/robots/dor_repo/accession/descriptive_metadata.rb
@@ -21,6 +21,7 @@ module Robots
             Dor::Services::Client.object(druid).refresh_metadata
           end
 
+          obj = Dor.find(druid) # reload object to get latest content
           raise "#{druid} descMetadata missing required fields (<title>)" if missing_required_fields?(obj.descMetadata)
         end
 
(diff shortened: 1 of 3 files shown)


ndushay:
@andrewjbtw w00t!   This took me entirely too long, but I do believe this was the problem.

lib/robots/dor_repo/accession/descriptive_metadata.rb
```diff
@@ -21,6 +21,7 @@ def perform(druid)
             Dor::Services::Client.object(druid).refresh_metadata
           end
 
+          obj = Dor.find(druid) # reload object to get latest content
```
mjgiarlo: `obj.reload` should work as well but it's NBD!

PR #397, opened 2019-09-16
Use the new update_status method in dor-workflow-client

jcoyne:


Gemfile.lock
```diff
@@ -430,4 +431,4 @@ DEPENDENCIES
   zeitwerk (~> 2.1)
 
 BUNDLED WITH
-   2.0.1
+   2.0.2
```
mjgiarlo: This line seems to making Travis angry.

PR #418, opened 2019-10-15
Reduce the reliance on Dor.find

jcoyne:
Instead call the dor-service-app API.  This will enable us to break the coupling to ActiveFedora and Fedora 3

### DIFF
--- a/lib/robots/dor_repo/accession/publish.rb
+++ b/lib/robots/dor_repo/accession/publish.rb
@@ -11,10 +11,10 @@ module Robots
         end
 
         def perform(druid)
-          obj = Dor.find(druid)
-          return unless obj.is_a?(Dor::Set) || obj.is_a?(Dor::Item)
+          object_client = Dor::Services::Client.object(druid)
+          obj = object_client.find
 
-          Dor::Services::Client.object(druid).publish
+          object_client.publish unless obj.is_a?(Cocina::Models::AdminPolicy)
         end
       end
     end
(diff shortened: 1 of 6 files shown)


lib/robots/dor_repo/accession/publish.rb
```diff
@@ -11,10 +11,10 @@ def initialize
         end
 
         def perform(druid)
-          obj = Dor.find(druid)
-          return unless obj.is_a?(Dor::Set) || obj.is_a?(Dor::Item)
+          object_client = Dor::Services::Client.object(druid)
+          obj = object_client.find
 
-          Dor::Services::Client.object(druid).publish
+          object_client.publish unless obj.is_a?(Cocina::Models::AdminPolicy)
```
mjgiarlo: I'm a fan of reducing `Dor.find` in the wild. My question about this line is where the type-checking should occur. Do you think users of the client should do this check, or should dor-services-client do this check internally?

jcoyne: I think we should continue to do it here (as we have done in the past).  This PR is about changing how we retrieve the model, not about the tests we perform on the mode.

mjgiarlo: OK, and this line is more or less equivalent to https://github.com/sul-dlss/common-accessioning/pull/418/files#diff-200aa82055a968339d85645db58e8fe7L15, so it's not a change in behavior.

PR #420, opened 2019-10-15
Run tests without having to spin up dependencies

jcoyne:


mjgiarlo:
@jcoyne also, and does this mean we can nuke `solr_conf/`?

PR #422, opened 2019-10-17
Add an experiment to see if this branch is used

jcoyne:


lib/dor/assembly/content_metadata.rb
```diff
@@ -77,6 +77,8 @@ def fnode_tuples(resource_type = '')
       def create_basic_content_metadata
         raise "Content metadata file #{Settings.assembly.cm_file_name} exists already for #{druid.id}" if content_metadata_exists?
 
+        Honeybadger.notify("How are we getting here? Shouldn't pre-assembly have generated contentMetadata.xml already?")
```
mjgiarlo: Do you want to include the druid in the message? Or is this sufficient for the experiment?

jcoyne: Yes, good call. I've added.

PR #435, opened 2019-10-31
Wait for publish jobs to finish

jcoyne:
## Why was this change made?

Publish jobs were made asynchronous as they were timing out. We should wait for them to complete.

n/a

### DIFF
--- a/lib/robots/dor_repo/accession/publish.rb
+++ b/lib/robots/dor_repo/accession/publish.rb
@@ -14,7 +14,12 @@ module Robots
           object_client = Dor::Services::Client.object(druid)
           obj = object_client.find
 
-          object_client.publish unless obj.is_a?(Cocina::Models::AdminPolicy)
+          return if obj.is_a?(Cocina::Models::AdminPolicy)
+
+          background_result_url = object_client.publish
+          result = object_client.async_result(url: background_result_url)
+          seconds = 2 * 60 * 60 # 2 hours of seconds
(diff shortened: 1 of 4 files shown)


lib/robots/dor_repo/accession/publish.rb
```diff
@@ -14,7 +14,12 @@ def perform(druid)
           object_client = Dor::Services::Client.object(druid)
           obj = object_client.find
 
-          object_client.publish unless obj.is_a?(Cocina::Models::AdminPolicy)
+          return if obj.is_a?(Cocina::Models::AdminPolicy)
+
+          background_result_url = Dor::Services::Client.object(druid).publish
```
mjgiarlo: I *think* you can reuse `object_client` here? So, `object_client.publish` instead.

lib/robots/dor_repo/accession/publish.rb
```diff
@@ -14,7 +14,12 @@ def perform(druid)
...
+          result = Dor::Services::Client::AsyncResult.new(url: background_result_url)
```
mjgiarlo: And I believe you can use the new interface here: `result = object_client.async_result(url: background_result_url)` (assuming that made it into a release)

jcoyne: that hasn't been released

mjgiarlo: now it has! 3.6.0

PR #436, opened 2019-10-31
Remove exception that obscures the source of the error

jcoyne:
## Why was this change made?
Previously Honeybadger was reporting any error in this method the same and obscuring the source.

n/a

### DIFF
--- a/lib/sdr_ingest_service.rb
+++ b/lib/sdr_ingest_service.rb
@@ -35,8 +35,6 @@ class SdrIngestService
 
     # start SDR preservation workflow
     Dor::Config.workflow.client.create_workflow_by_name(druid, 'preservationIngestWF')
-  rescue Exception => e
-    raise Dor::Exception, "Error exporting new object version to bag for #{dor_item.pid}: #{e.message}"
   end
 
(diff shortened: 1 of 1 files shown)


lib/sdr_ingest_service.rb
```diff
@@ -35,8 +35,6 @@ def self.transfer(dor_item)
 
     # start SDR preservation workflow
     Dor::Config.workflow.client.create_workflow_by_name(druid, 'preservationIngestWF')
-  rescue Exception => e
-    raise Dor::Exception, "Error exporting new object version to bag for #{dor_item.pid}: #{e.message}"
```
mjgiarlo: This change also removes the `Error exporting new object version to bag` string from the exception. Do we think this is not a helpful breadcrumb?

jcoyne: I don't think so.  I think we know if there is an exception here that's the consequence, but we don't know how to fix it.

mjgiarlo: Rescuing once for the entire method does make it harder to tell what's erroring. Going to approve this.

PR #438, opened 2019-10-31
[Need PO approval] Allow dor-services-app to use callbacks rather than polling

jcoyne:
## Why was this change made?

The polling operations can timeout and are not possible to reset using argo.
Depends on https://github.com/sul-dlss/dor-services-app/pull/490
 and https://github.com/sul-dlss/workflow-server-rails/pull/271

n/a


andrewjbtw:
This looks good to me on stage. I have one more test (a large video file) running through that's delayed at technical metadata but the others have succeeded.

PR #444, opened 2019-11-08
Remove unnecessary object parameter from DatastreamBuilder

jcoyne:
## Why was this change made?

There's no need to pass object as an argument, the datastream object has all the information this class needs.

n/a


jcoyne:
@mjgiarlo it's no longer broken.

PR #455, opened 2019-11-25
Call the SdrIngestService in dor-services-app

jcoyne:
## Why was this change made?

This reduces coupling with Fedora and the /dor/export filesystem.

Depends on https://github.com/sul-dlss/dor-services-app/pull/506 and 
https://github.com/sul-dlss/workflow-server-rails/pull/293

n/a


ndushay:
@jcoyne @mjgiarlo (tagging Mike as part of SDR evo brain trust)

I would love to have a broader conversation about this before porting this functionality as your PRs are proposing.   I'm splatting some thoughts here to try to clarify my own thinking and also to give you a window into my ideas/concerns.

Where do we want interactions with preservation to be limited to?  Putting them in dor-services-app feels uncomfortable to me;  in a perfect world, DSA is about interacting with DOR and not with preservation or workflows.

I think these are the current conceptual steps in production now:

- accessioning robots get to point where new material needs to be preserved
- accessioning sdr-ingest-transfer step:
  - creates bag for preservation service
    - gets new version number by getting signature catalog from existing Moab, or creates new one
    - writes out specific metadata files from DOR to "/dor/workspace" (desc, identity, prov, rels, version)
    - verifies the version from versionMetadata.xml in DOR aligns with the new version number
    - creates new metadata and manifest files needed in to-be-updated Moab, using files in "/dor/workspace", new files created using moab-versioning gem
    - creates a BagIt bag, leveraging moab-versioning gem, for preservation robots
  - notifies preservation robots to start by creating preservationIngestWF
- common accessioning now paused

- preservation robots:
  - copies BagIt bag from common-accessioning location to the correct deposit location for the Moab
  - validates, verifies stuff
  - creates or adds a version to Moab object from bag
  - updates prescat db
  - removes deposit bag created by transfer-object robot
  - transfers control back to accessioning by marking the next step in the common-accessioning WF completed

- accessioning may now finish
  - reset-workspace WF step
  - end-accession WF step

Problems with current production (?):

- I gather we would like a better mechanism for transferring control from accessioningWF to preservationIngestWF and back to accessioningWF.   That is grand!  It can be separated from the other refactoring. 

- Is there some hope of improving/simplifying our cleanup stuff via this?  Also would be grand ... but could be done separately?

Here's the meat of my specific concerns:

- making the deposit BagIt bag for the Moab seems to require:
  - info/files from preservation 
    - I propose preservationAPI, accessed with preservation-client from common-accessioning (change:  all via preservation-client, no direct access to Moab)
      - I just wrote the code to get signatureCatalog file with preservation-client, not via dor-services-client talking to dor-services-app to make call to sdr-services-app.
    - I believe we want to avoid having dor-services-app need to use preservation-client if possible (clearer separation of purpose: keep dor-services-app focused on DOR)
    - When we migrated "current_version" lookups to preservation-client, we made a point of directly hooking up the callers with preservation-client directly and deprecating the method in DSA.  I did the same for the checksums lookup, once the preservation-client gem was created.
  - files in "/dor/workspace"
    - I propose common-accessioning, not dor-services (change: none - keep as is in prod now).  

- we should make sure the API at preservation catalog gives all the needed access to preserved objects.  I think sdr-ingest-service code and possibly one or two other places (e.g. the shelving service for content files) are the only remaining spots that need this access.

jcoyne:
@ndushay I'm just considering this step, which is to move the "preserve an object" to an API as the first step.  I think we have lots of room for further improvement, but this begins to align the design with the API oriented design that we created in the COCINA workcycle.

mjgiarlo:
@jcoyne @ndushay I've added this as an agenda item to Tuesday's @sul-dlss/infrastructure-team dev planning meeting so we can discuss the architecture going forward.

justinlittman:
So just to fess up that I merged this without seeing the discussion (it was below the fold).

ndushay:
I don't think this is in dor-services-app yet -- is it?

mjgiarlo:
@ndushay this work has been reverted, btw.

PR #456, opened 2019-11-26
Update descMetadata using an API

jcoyne:
## Why was this change made?

This decouples the robot from Fedora and switches to an API based interface.

n/a

### DIFF
--- a/lib/robots/dor_repo/accession/descriptive_metadata.rb
+++ b/lib/robots/dor_repo/accession/descriptive_metadata.rb
@@ -12,22 +12,20 @@ module Robots
         end
 
         def perform(druid)
-          obj = Dor.find(druid)
-          builder = DatastreamBuilder.new(datastream: obj.descMetadata)
-          builder.build do |_ds|
-            # If there's no file on disk that's newer than the datastream and
-            # the datastream has never been populated, use Symphony:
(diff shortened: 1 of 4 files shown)


lib/robots/dor_repo/accession/descriptive_metadata.rb
```diff
@@ -12,22 +12,20 @@ def initialize
...
-          obj = Dor.find(druid)
-          builder = DatastreamBuilder.new(datastream: obj.descMetadata)
-          builder.build do |_ds|
-            # If there's no file on disk that's newer than the datastream and
-            # the datastream has never been populated, use Symphony:
-            Dor::Services::Client.object(druid).refresh_metadata
-          end
-
-          obj.reload # get latest content before check
-          raise "#{druid} descMetadata missing required fields (<title>)" if missing_required_fields?(obj.descMetadata)
```
ndushay: very late to the party here -- did this check for missing descMetadata req fields get moved to dor-services-client or somewhere?

jcoyne: It's in dor-services-app now.

jcoyne: See https://github.com/sul-dlss/dor-services-app/pull/507

PR #457, opened 2019-11-26
Refactor TechnicalMetadataService to use Ruby conventions

jcoyne:
## Why was this change made?

This code was challenging to read as it didn't use common conventions.

n/a


lib/technical_metadata_service.rb
```diff
@@ -14,36 +14,47 @@ class TechnicalMetadataService
...
+  end
+
+  def add_update_technical_metadata
+    diff = content_group_diff
+    deltas = diff.file_deltas
+    old_techmd = old_technical_metadata
+    new_techmd = new_technical_metadata(deltas)
+
+    # this is version 1 or previous technical metadata was not saved
+    return store(dor_item, new_techmd) if old_techmd.nil?
```
mjgiarlo: Isn't `#store`'s arity 1 (per line 44)? I suspect we need to remove `dor_item` here which is now an attr.

mjgiarlo: It's a bit alarming that if :point_up: is right, we're not seeing any test failures. But maybe the current code has that problem too?

jcoyne: Thank you for your careful review. I'll add a test for that.

mjgiarlo: and thank you for continuing to weed and prune in these codebases

jcoyne: Fixed and added a test.

PR #460, opened 2019-12-03
Inline Checksumable as it's only used by one consumer

jcoyne:
## Why was this change made?
Checksumable was only used by one consumer, so there's no need to have it on Assembly::Item.
Also added testing of the checksum compute code.

n/a


lib/dor/assembly/content_metadata.rb
```diff
@@ -30,12 +30,14 @@ def stub_content_metadata_exists?
         File.exist?(stub_cm_file_name)
       end
 
+      # rubocop:disable Naming/MemoizedInstanceVariableName
       def load_content_metadata
         # Loads content metadata XML into a Nokogiri document.
         raise "Content metadata file #{Settings.assembly.cm_file_name} not found for #{druid.id} in any of the root directories: #{@root_dir.join(',')}" unless content_metadata_exists?
 
-        @cm = Nokogiri.XML(File.open(cm_file_name)) { |conf| conf.default_xml.noblanks }
+        @cm ||= Nokogiri.XML(File.open(cm_file_name)) { |conf| conf.default_xml.noblanks }
```
mjgiarlo: Can you say more about how this change fits in this PR? Is it an opportunistic performance bump?

lib/robots/dor_repo/assembly/checksum_compute.rb
```diff
@@ -9,7 +9,59 @@ def initialize(opts = {})
         end
 
         def perform(druid)
-          with_item(druid, &:compute_checksums)
+          with_item(druid) do |assembly_item|
```
mjgiarlo: Glad to see this inlined. :+1:

lib/dor/assembly/content_metadata.rb
jcoyne: It allows us to make expectations on the values returned from this.  This ensures that every call on the instance returns the same values.

PR #461, opened 2019-12-03
Extract PathFinder from Assembly::Item

jcoyne:
## Why was this change made?

There is no reason to have a module that is only used once. Make it a class instead

n/a

### DIFF
--- a/lib/dor/assembly/path_finder.rb
+++ b/lib/dor/assembly/path_finder.rb
@@ -2,8 +2,16 @@
 
 module Dor
   module Assembly
-    module Findable
-      attr_reader :folder_style
+    # This finds files associated with a Druid
+    class PathFinder
+      # @param [String] The non-namespaced druid identifier.
+      def initialize(druid_id:)
+        @druid_id = druid_id
+      end
+
+      def check_for_path
+        raise "Path to object #{druid_id} not found in any of the root directories: #{root_dir.join(',')}" if path_to_object.nil?
+      end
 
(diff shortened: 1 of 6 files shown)


lib/dor/assembly/path_finder.rb
```diff
@@ -47,15 +45,33 @@ def path_to_content_file(filename)
...
+
+      attr_reader :folder_style
+
+      def root_dir
+        Array(Settings.assembly.root_dir)
+      end
+
+      # new style path, e.g. aa/111/bb/2222/aa111bb2222
+      def druid_tree_path(root_dir)
+        DruidTools::Druid.new(@druid.id, root_dir).path
```
mjgiarlo: Worth considering adding an attr for `@druid`?

lib/dor/assembly/path_finder.rb
```diff
@@ -47,15 +45,33 @@ def path_to_content_file(filename)
...
+      end
+
+      # old style path, e.g. aa/111/bb/2222
+      def old_druid_tree_path(root_dir)
+        File.dirname druid_tree_path(root_dir)
+      end
+
+      # path to a metadata folder, defaults to new (aa/111/bb/2222/aa111bb2222/metadata), but could also be old style (aa/111/bb/2222)
+      def path_to_metadata_folder
+        @folder_style == :old ? path_to_object : File.join(path_to_object, 'metadata')
```
mjgiarlo: Worth considering adding an attr for `@folder_style`?

jcoyne:
@mjgiarlo I've added the recommended attr_readers.

PR #465, opened 2019-12-07
 technical_metadata_service uses preservation-client instead of sdr-services-app via dor-services-client

ndushay:
~~must not be deployed until sul-dlss/preservation_catalog/pull/1250 is deployed.~~ <-- done.

## Why was this change made?

To get rid of  `cm-inv-diff` and `metadata` API endpoints in sdr-services-app.  We use preservation-client methods instead.

(Note that this PR does not include calls in SdrIngestService as that is being ported to dor-services-app with various WIP PRs)

n/a


jcoyne:
Would this require any firewall changes?

ndushay:
I already checked and the common-accessioning boxes can reach preservation-catalog and the shared_configs PRs are already merged.  :-)

config/settings.yml
```diff
@@ -19,7 +19,9 @@ release:
   max_sleep_seconds: 1  # max sleep seconds between tries
   base_sleep_seconds: 1 # base sleep seconds between tries
 
-
 sdr:
   local_workspace_root: /dor/workspace
   local_export_home: /dor/export
+
+preservation_catalog:
+  url: 'https://example.org/prescat'
```
jcoyne: Do we need this in shared_configs before merging/deploying this?

ndushay: see my comment above - it's already there.

PR #474, opened 2019-12-11
TechnicalMetadataService copes PROPERLY with no technical metadata from preservation

ndushay:
## Why was this change made?

b/c I goofed in the previous PR #472 to fix https://app.honeybadger.io/projects/52894/faults/58394059.

I believe the real problem was that the regex didn't accommodate multiple lines;  it is also a bit worrying that the exception from pres-client isn't a NotFound, but I have a diff ticket for that.

n/a


jmartin-sul:
merge at your leisure

ndushay:
fix deployed, problem well and truly fixed now.

PR #476, opened 2019-12-11
Revert "Revert "Call the SdrIngestService in dor-services-app""

mjgiarlo:
Reverts sul-dlss/common-accessioning#475


mjgiarlo:
@jcoyne Feel free to make this a real PR and go forward with it once you're ready to coordinate the deploys (w/ dor-services-app and workflow-server-rails).

PR #477, opened 2019-12-11
preservation-client v1.0.0 to get technicalMetadata.xml

ndushay:
## Why was this change made?

preservation-client v1.0.0 throws a Preservation::Client::NotFoundError when an API call to preservation-catalog returns 404.  This simplifies code here that traps for that error.

n/a

### DIFF
--- a/lib/technical_metadata_service.rb
+++ b/lib/technical_metadata_service.rb
@@ -76,7 +76,7 @@ class TechnicalMetadataService
     dor_technical_metadata
   end
 
-  # @return [String] The technicalMetadata datastream from the previous version of the digital object (fetched from preeservation)
+  # @return [String] The technicalMetadata datastream from the previous version of the digital object (fetched from preservation)
   #   The data is updated to the latest format.
   def preservation_technical_metadata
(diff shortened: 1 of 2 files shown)


lib/technical_metadata_service.rb
```diff
@@ -103,12 +103,8 @@ def dor_technical_metadata
   # @return [String] The datastream contents from the previous version of the digital object (fetched from preservation)
```
mjgiarlo: Would you mind amending the YARD doc for `@return` here to reflect that (and why) `nil`s are returned?

ndushay: can do.

ndushay:
I already wrote the test for it in a previous PR -- the test raises a NotFoundError already.

lib/technical_metadata_service.rb
ndushay: done.

mjgiarlo:
> I already wrote the test for it in a previous PR -- the test raises a NotFoundError already.

Aha! I missed that. This looked like a change in behavior but it's really not (it's just a change in how we detect the "not found" state). Thanks! Approving.

PR #478, opened 2019-12-12
Don't refresh metadata if the MODS already exists

jcoyne:
## Why was this change made?
This fixes a problem introduced in #456 where the MODS is being overwritten with the MARC data


peetucket:
This seems reasonable but its also not clear to me why the changes in https://github.com/sul-dlss/common-accessioning/pull/456 would be different.  I am not seeing any checks like this in that previous version of the code before it calls `.refresh_metadata`

ndushay:
I'm going to merge this and deploy it ASAP b/c it's a blocker for prod.

PR #487, opened 2020-01-06
Extract a member service from Release::Item

jcoyne:
## Why was this change made?

A class should have a single responsibility.  The Release::Item class had several responsibilities, this extracts the responsibility for finding collection members to its own class

n/a

### DIFF
--- a/spec/lib/dor/release/item_spec.rb
+++ b/spec/lib/dor/release/item_spec.rb
@@ -5,15 +5,11 @@ require 'spec_helper'
 RSpec.describe Dor::Release::Item do
   before do
     @druid = 'oo000oo0001'
-    @item = described_class.new(druid: @druid, skip_heartbeat: true) # skip heartbeat check for dor-fetcher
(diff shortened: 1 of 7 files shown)


spec/lib/dor/release/item_spec.rb
```diff
@@ -32,22 +28,6 @@
...
-  end
-
-  it 'gets the right value for item_members' do
-    expect(@item.item_members).to eq @response['items']
-  end
-
-  it 'gets the right value for sub_collections' do
-    expect(@item.sub_collections).to eq @response['sets'] + @response['collections']
-  end
-
```
ndushay: it looks like these specs didn't migrate anywhere ... would it make sense to have this be the content of a spec for the new MemberService class?

jcoyne: Right, because we have other tests that provide 100% coverage of MemberService: https://coveralls.io/builds/28005421/source?filename=lib/dor/release/member_service.rb

PR #493, opened 2020-01-07
Refactor to put the updating marc code close to where it is used

jcoyne:
## Why was this change made?
Removes unnecessary indirection on how the update_marc code is called.

n/a

### DIFF
--- a/spec/robots/release/update_marc_spec.rb
+++ b/spec/robots/release/update_marc_spec.rb
@@ -2,16 +2,13 @@
 
 require 'spec_helper'
 
-describe Robots::DorRepo::Release::UpdateMarc do
-  before do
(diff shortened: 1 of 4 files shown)


spec/robots/release/update_marc_spec.rb
```diff
@@ -2,16 +2,13 @@
...
-  before do
-    @druid = 'aa222cc3333'
-    @work_item = instance_double(Dor::Item)
-    @r = described_class.new
-  end
+RSpec.describe Robots::DorRepo::Release::UpdateMarc do
+  let(:druid) { 'aa222cc3333' }
+  let(:robot) { described_class.new }
 
   it 'calls the update marc record method' do
```
ndushay: this description is no longer accurate.

PR #513, opened 2020-01-21
For robots, when creating a new workflow sets lane id from calling wo…

### LINKED ISSUE #512: Pass on lane id when creating new workflow
So that the child workflow has the same lane id as the parent.

justinlittman:
…rkflow step.

closes #512

## Why was this change made?
So that lanes are passed to newly spawned workflows.

No.


lib/robots/dor_repo/etd_submit/base.rb
```diff
@@ -1,15 +0,0 @@
-# frozen_string_literal: true
```
jcoyne: Yay!

PR #521, opened 2020-01-27
[HOLD] Decouple from Solr

jcoyne:
## Why was this change made?

The common-accessioning doesn't need to index because the dor-indexing-app should be responsible for that. Previously we've been indexing more than one time for each change, which causes undue load on the system.

This is being held until we can test on stage.

n/a


jcoyne:
Solr is no longer a dependency.

PR #522, opened 2020-01-27
Use the cocina endpoint to determine object_type

jcoyne:
## Why was this change made?

This avoids a call to Fedora

n/a

### DIFF
--- a/lib/dor/release/item.rb
+++ b/lib/dor/release/item.rb
@@ -17,12 +17,12 @@ module Dor
         @object ||= Dor.find(@druid)
       end
 
-      def object_type
-        unless @obj_type
-          obj_type = object.identityMetadata.objectType
-          @obj_type = (obj_type.nil? ? 'unknown' : obj_type.first)
-        end
-        @obj_type.downcase.strip
+      def cocina_model
+        @cocina_model ||= Dor::Services::Client.object(@druid).find
+      end
+
+      def collection?
+        cocina_model.is_a?(Cocina::Models::Collection)
       end
     end
   end
--- a/lib/robots/dor_repo/release/release_members.rb
+++ b/lib/robots/dor_repo/release/release_members.rb
@@ -18,12 +18,9 @@ module Robots
 
           item = Dor::Release::Item.new druid: druid
 
-          case item.object_type
-          when 'collection', 'set' # this is a collection or set
-            publish_collection(druid, item)
-          else # this is not a collection of set
-            LyberCore::Log.debug "...this is a #{item.object_type}, NOOP"
-          end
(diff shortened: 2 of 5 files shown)


lib/dor/release/item.rb
```diff
@@ -17,12 +17,12 @@ def object
...
-        unless @obj_type
-          obj_type = object.identityMetadata.objectType
-          @obj_type = (obj_type.nil? ? 'unknown' : obj_type.first)
-        end
-        @obj_type.downcase.strip
+      def cocina_model
+        @cocina_model ||= Dor::Services::Client.object(@druid).find
+      end
+
+      def collection?
```
mjgiarlo: Not a blocking comment: might we later move this method, and related methods (*e.g.*, `#dro?`), into the `cocina-models` gem?

lib/robots/dor_repo/release/release_members.rb
```diff
@@ -18,12 +18,9 @@ def perform(druid)
 
           item = Dor::Release::Item.new druid: druid
 
-          case item.object_type
-          when 'collection', 'set' # this is a collection or set
-            publish_collection(druid, item)
-          else # this is not a collection of set
-            LyberCore::Log.debug "...this is a #{item.object_type}, NOOP"
-          end
+          return unless item.collection?
```
mjgiarlo: I assume we're OK dropping the debug log entry above. It doesn't look very useful nor does it provide actionable information.

lib/robots/dor_repo/release/release_members.rb
```diff
@@ -35,24 +32,17 @@ def publish_collection(druid, item)
           #   if at least one of the targets is *not* what=self, we will do it
           tag_service = Dor::ReleaseTagService.for(item.object)
           release_tags = tag_service.newest_release_tag(tag_service.release_tags) # get the latest release tag for each target
-          if release_tags.collect { |_k, v| v['what'] == 'self' }.include?(false) # if there are any *non* what=self release tags in any targets, go ahead and add the workflow to the items
-            add_workflow_to_members(member_service, item.object_type)
-          else # all of the latest release tags are what=self or there are no release tags, so skip
-            LyberCore::Log.debug "...all release tags are what=self for #{item.object_type}; skipping member workflows"
-          end
+          # if there are any *non* what=self release tags in any targets, go ahead and add the workflow to the items
+          add_workflow_to_members(member_service) if release_tags.collect { |_k, v| v['what'] == 'self' }.include?(false)
```
mjgiarlo: I suspect there's a more streamlined way to detect if any `'what'` values are not `'self'` but this is not related to removing the debug call so `/shrug`

PR #525, opened 2020-01-27
Refactor RightsMetadata robot to not depend on DatastreamBuilder

jcoyne:
## Why was this change made?

We're trying to move to using the API methods for updating metadata rather than directly communicating with Fedora

n/a

### DIFF
--- a/lib/robots/dor_repo/accession/rights_metadata.rb
+++ b/lib/robots/dor_repo/accession/rights_metadata.rb
@@ -10,17 +10,34 @@ module Robots
         end
 
         def perform(druid)
-          obj = Dor.find(druid)
-          builder = DatastreamBuilder.new(datastream: obj.rightsMetadata)
+          object = DruidTools::Druid.new(druid, Dor::Config.stacks.local_workspace_root)
+          path = object.find_metadata('rightsMetadata.xml')
+          object_client = Dor::Services::Client.object(druid)
+          if path
+            object_client.metadata.legacy_update(
+              rights: {
+                updated: File.mtime(path),
+                content: File.read(path)
+              }
+            )
+          elsif has_no_rights_metadata?(druid)
+            Honeybadger.notify("I don't think this ever happens because rights is created when registering. This is an experiment")
 
-          builder.build do |datastream|
-            build_datastream(obj, datastream)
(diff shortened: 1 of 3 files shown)


lib/robots/dor_repo/accession/rights_metadata.rb
```diff
@@ -10,17 +10,34 @@ def initialize
...
-            build_datastream(obj, datastream)
+            # TODO: Fetch admin_policy_object without involving Fedora 3 concepts
+            apo = Dor.find(druid).admin_policy_object
+            object_client.metadata.legacy_update(
+              rights: {
+                updated: Time.now,
+                content: apo.defaultObjectRights.content
+              }
+            )
           end
```
justinlittman: Is there an else case that needs to be handled? (rightsMetadata.xml not on disk and object has rights metadata)

lib/robots/dor_repo/accession/rights_metadata.rb
```diff
@@ -10,17 +10,34 @@ def initialize
...
         end
 
-        def build_datastream(obj, datastream)
-          datastream.dsLabel = 'Rights Metadata'
-          datastream.ng_xml = obj.admin_policy_object.defaultObjectRights.ng_xml.clone
+        # TODO: for now we're looking for the presence of the datastream, but eventually
+        # we need to do this without involving Fedora 3 concepts
+        def has_rights_metadata?(druid)
+          !Dor.find(druid).rightsMetadata.new?
```
justinlittman: The negation here and L23 makes code a little difficult to follow. Would be a bit easier if made this `has_no_rights_metadata`.

lib/robots/dor_repo/accession/rights_metadata.rb
jcoyne: I don't think so, I think the two legitimate cases are "there is rightsMetadata.xml on disk" and "do nothing" (because registration gives every object rightsMetadata.xml).

lib/robots/dor_repo/accession/rights_metadata.rb
jcoyne: That sounds good. Change made.

PR #527, opened 2020-01-27
Refactor the technical metadata robot not to use DatastreamBuilder

jcoyne:
## Why was this change made?
Instead use the API methods. This positions the project to remove the Fedora dependency

n/a

### DIFF
--- a/lib/robots/dor_repo/accession/technical_metadata.rb
+++ b/lib/robots/dor_repo/accession/technical_metadata.rb
@@ -13,10 +13,29 @@ module Robots
           obj = Dor.find(druid)
           return unless obj.is_a?(Dor::Item)
 
-          builder = DatastreamBuilder.new(datastream: obj.technicalMetadata,
-                                          force: true)
-          builder.build do |_datastream|
-            TechnicalMetadataService.add_update_technical_metadata(obj)
+          object = DruidTools::Druid.new(druid, Dor::Config.stacks.local_workspace_root)
+          path = object.find_metadata('technicalMetadata.xml')
+          object_client = Dor::Services::Client.object(druid)
+
+          if path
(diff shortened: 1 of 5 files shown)


lib/robots/dor_repo/accession/technical_metadata.rb
```diff
@@ -13,10 +13,27 @@ def perform(druid)
...
 
-          builder = DatastreamBuilder.new(datastream: obj.technicalMetadata,
-                                          force: true)
-          builder.build do |_datastream|
-            TechnicalMetadataService.add_update_technical_metadata(obj)
+          object = DruidTools::Druid.new(druid, Dor::Config.stacks.local_workspace_root)
+          path = object.find_metadata('technicalMetadata.xml')
+          object_client = Dor::Services::Client.object(druid)
+
+          if path
```
justinlittman: Some comments would be really helpful.

lib/technical_metadata_service.rb
```diff
@@ -30,13 +30,12 @@ def add_update_technical_metadata
     old_techmd = old_technical_metadata
     new_techmd = new_technical_metadata(deltas)
     # this is version 1 or previous technical metadata was not saved
-    return store(new_techmd) if old_techmd.nil?
+    return new_techmd if old_techmd.nil?
```
justinlittman: Docs above need to change if method signature is changing.

jcoyne: Thanks! I've updated the return value for `.add_update_technical_metadata`

lib/robots/dor_repo/accession/technical_metadata.rb
jcoyne: Added.

PR #529, opened 2020-01-28
Do dependency injection for TechnicalMetadataService

jcoyne:
## Why was this change made?

This removes the TechnicalMetadataService's dependency on Fedora

n/a

### DIFF
--- a/lib/technical_metadata_service.rb
+++ b/lib/technical_metadata_service.rb
@@ -11,15 +11,15 @@ require 'jhove_service'
 # Switching to a more granular data model that has file metadata separate from
 # the Work metadata will allow us to simplify this greatly.
 class TechnicalMetadataService
-  # @param [Dor::Item] dor_item The DOR item being processed by the technical metadata robot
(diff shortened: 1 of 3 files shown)


lib/technical_metadata_service.rb
```diff
@@ -11,14 +11,17 @@
 # Switching to a more granular data model that has file metadata separate from
 # the Work metadata will allow us to simplify this greatly.
 class TechnicalMetadataService
-  # @param [Dor::Item] dor_item The DOR item being processed by the technical metadata robot
+  # @param [String] content_metadata The xml representation of the content metadata from DOR
+  # @param [String] tech_metadata The xml representation of tech metadata from DOR
```
mjgiarlo: `tech_metadata` is expected to be a string but may also be `nil` if the techMD datastream is `#new?`, right? Worth reflecting in the YARD documentation that this `@param` may be either a string or a `nil`?

jcoyne: Sure, I can make that change.

mjgiarlo: Thanks, @jcoyne.

PR #531, opened 2020-01-28
Update provenance without hitting Fedora

### LINKED ISSUE #528: provenance_metadata robot should not call Fedora
Instead use the legacy metadata update method.

jcoyne:
## Why was this change made?

So that common-accessioning will no longer depend on Fedora

Fixes #528 

n/a

### DIFF
--- a/lib/robots/dor_repo/accession/provenance_metadata.rb
+++ b/lib/robots/dor_repo/accession/provenance_metadata.rb
@@ -9,34 +9,32 @@ module Robots
         end
 
         def perform(druid)
-          obj = Dor.find(druid)
-          build_datastream(obj, 'accessionWF', 'DOR Common Accessioning completed')
+          workflow_provenance = create_workflow_provenance(druid)
+          object_client = Dor::Services::Client.object(druid)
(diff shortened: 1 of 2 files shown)


lib/robots/dor_repo/accession/provenance_metadata.rb
```diff
@@ -9,34 +9,32 @@ def initialize
         end
 
         def perform(druid)
-          obj = Dor.find(druid)
-          build_datastream(obj, 'accessionWF', 'DOR Common Accessioning completed')
+          workflow_provenance = create_workflow_provenance(druid)
+          object_client = Dor::Services::Client.object(druid)
+          object_client.metadata.legacy_update(
+            provenance: {
+              updated: Time.now,
```
mjgiarlo: I'm not sure it matters, but do we care if the timestamps in the `updated` field here and the WF XML event node differ?

jcoyne: Not important because, this is only relevant to dor-services-app when it sees if it's newer than what is stored.

PR #532, opened 2020-01-28
Encapsulate the Fedora object just to where it's needed

jcoyne:
## Why was this change made?

This will assist in helping us break the dependency on Fedora

n/a

### DIFF
--- a/lib/robots/dor_repo/accession/technical_metadata.rb
+++ b/lib/robots/dor_repo/accession/technical_metadata.rb
@@ -10,12 +10,14 @@ module Robots
         end
 
         def perform(druid)
-          obj = Dor.find(druid)
-          return unless obj.is_a?(Dor::Item)
+          object_client = Dor::Services::Client.object(druid)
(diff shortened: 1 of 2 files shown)


lib/robots/dor_repo/accession/technical_metadata.rb
```diff
@@ -10,12 +10,14 @@ def initialize
         end
 
         def perform(druid)
-          obj = Dor.find(druid)
-          return unless obj.is_a?(Dor::Item)
+          object_client = Dor::Services::Client.object(druid)
+          obj = object_client.find
+
+          # non-items don't generate contentMetadata
+          return unless obj.is_a?(Cocina::Models::DRO)
```
mjgiarlo: Given https://github.com/sul-dlss/common-accessioning/pull/537, are you open to changing this to `return unless obj.dro?` ?

jcoyne: Sure.

PR #543, opened 2020-01-29
Don't run techMetadata or shelve when there are no files

### LINKED ISSUE #540: If an object has no files, don't run technicalMetadata or shelve robots
Presently if there are no files (no contentMetadata.xml) then it tries to get content_md_diff from preservation with nil which causes preservation catalog to raise an error.

This may have behaved differently when we had sdr-services-app in the stack.

we have decided that it's clearer if the robots skip technical_metadata and shelve if there are no files.

jcoyne:
## Why was this change made?

Fixes #540

n/a

### DIFF
--- a/spec/robots/accession/shelve_spec.rb
+++ b/spec/robots/accession/shelve_spec.rb
@@ -15,17 +15,16 @@ RSpec.describe Robots::DorRepo::Accession::Shelve do
 
     let(:object_client) { instance_double(Dor::Services::Client::Object, find: object, shelve: nil) }
 
+    before do
+      allow(robot.workflow_service).to receive(:update_status)
+    end
+
     context 'when called on a non-item' do
       let(:object) do
         Cocina::Models::Collection.new(externalIdentifier: '123',
-                                       type: Cocina::Models::Collection::TYPES.first,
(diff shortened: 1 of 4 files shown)


jcoyne:
I tested this on stage by registering an object in argo and then using argo to start the accessioningWF.  This previously triggered the bug in #540, but with this PR it works:
Screen Shot 2020-01-29 at 10 00 15 PM

spec/robots/accession/shelve_spec.rb
```diff
@@ -41,19 +40,48 @@
...
                                 label: 'my repository object',
                                 version: 1)
       end
 
-      context "when it's successful" do
-        it 'shelves the item' do
-          perform
-          expect(object_client).to have_received(:shelve)
-        end
+      it 'shelves the item' do
```
justinlittman: does not shelve the item?

spec/robots/accession/technical_metadata_spec.rb
```diff
@@ -24,11 +24,21 @@
     context 'on an item' do
       let(:dor_object) { Dor::Item.new(pid: druid) }
 
+      let(:file_set) do
```
justinlittman: Need a test when no files?

PR #546, opened 2020-01-31
Release sub-collections using tags from cocina

### LINKED ISSUE #538: Remove coupling to dor-services in release_members robot
https://github.com/sul-dlss/common-accessioning/blob/master/lib/robots/dor_repo/release/release_members.rb#L28-L33

This might have to be a new api in dor-services-app, which this robot calls via dor-services-client

jcoyne:
## Why was this change made?

This removes one more dependency on Fedora 3.

Fixes #538

n/a


jcoyne:
@ndushay it's doing the same work, but without looking directly at Fedora.  It's looking at the API provided  by dor-services-app instead.

PR #554, opened 2020-02-04
Use a local variable instead of an ivar

jcoyne:
## Why was this change made?
The ivar is not defined

n/a


jcoyne:
Dup of #553

PR #558, opened 2020-02-06
 Test for stub_content_metadata

### LINKED ISSUE #518: Write test for convert_stub_content_metadata
There are no tests for this method.

jermnelson:
## Why was this change made?
Fixes #518 

n/a


spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
```diff
@@ -47,5 +69,12 @@
         end
       end
     end
+
+    context 'when convert_stub_content_metadata method is called' do
+      it 'generates stub content metadata' do
```
jcoyne: Another way this can be written is to use the subject.  See http://www.betterspecs.org/#subject

```suggestion
      subject { item.convert_stub_content_metadata }
      it { is_expected.to be_equivalent_to xml_doc }
```

PR #560, opened 2020-02-06
Avoid LyberCore::Robot.step_to_classname

jcoyne:
## Why was this change made?
It's deprecated because it's a lot more indirection than is required

### DIFF
--- a/bin/run_robot
+++ b/bin/run_robot
@@ -5,12 +5,12 @@
 #
 # To run With Options
 # Options must be placed AFTER workflow and robot name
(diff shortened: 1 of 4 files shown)


bin/run_robot
```diff
@@ -1,16 +1,17 @@
+
```
mjgiarlo: phantom keyboard strikes again

mjgiarlo: I suspect you'll need to wipe out this line before merging.

PR #561, opened 2020-02-06
Lyber core 6

jcoyne:
## Why was this change made?
This helps decouple from dor-services.

n/a


bin/run_robot
```diff
@@ -1,16 +1,17 @@
+
```
mjgiarlo: :exclamation:

Issue #428, opened 2019-10-29
Merge Accessioning end-accession and Dissemination workflow.


jcoyne:
https://github.com/sul-dlss/common-accessioning/blob/master/lib/robots/dor_repo/accession/end_accession.rb#L19-L20

https://github.com/sul-dlss/common-accessioning/blob/master/lib/robots/dor_repo/dissemination/cleanup.rb#L12

The end-accession step calls dissemination workflow, which only has one step, which is to call cleanup.  The could be merged into reset-workspace.

mjgiarlo:
Need to talk with @andrewjbtw about whether this is a useful or duplicative practice. If the latter, consider in-scope for GBooks SDR Evolution WC.

jcoyne:
Ensure no tie to web-archiving

jcoyne:
Talked with @andrewjbtw and he said, that he can't see why this needs to exist on its own.

ndushay:
the `end-accession` step calls `startSpecialDissemination`, which signals to was_robots_suite dissemination robot to trigger (and then call either seed or crawl dissemination).   So the trick will be to ensure that if we combine the steps in some way, the ordering doesn't fooch the web-archiving stuff (by removing things too soon?) or the last step somehow.  

Justin and I quickly looked over the code and believe there is no conflict but it wouldn't hurt to double check.

ndushay:
do we need to make any changes to any workflows expectations and/or will there be any problems for objects stuck in processing the old way when we go to the new way?

mjgiarlo:
Do you have thoughts on @ndushay's question above, @jcoyne?

jcoyne:
Shouldn't be a problem as there's nothing stopped in those steps.

justinlittman:
@ndushay @mjgiarlo @jcoyne Anything else need to be done for this ticket other than https://github.com/sul-dlss/common-accessioning/pull/563?

I reviewed the mention of dissemination WF in other repos and don't see anything else that needs to be changed.

mjgiarlo:
@justinlittman Not that I am aware of.

PR #563, opened 2020-02-07
Merge Accessioning end-accession and Dissemination workflow.

justinlittman:
refs #428

## Why was this change made?
The DisseminationWF doesn't do anything.

No.


spec/robots/accession/end_accession_spec.rb
```diff
@@ -24,9 +24,10 @@
   let(:apo_id) { 'druid:mx121xx1234' }
   let(:process) { instance_double(Dor::Workflow::Response::Process, lane_id: 'default') }
   let(:workflow_client) { instance_double(Dor::Workflow::Client, create_workflow_by_name: nil, process: process) }
```
jcoyne: Can `create_workflow_by_name` be removed here?

justinlittman: No, see L60.

PR #577, opened 2020-02-20
Switches to invoking techmd service.

### LINKED ISSUE #575: [HOLD] Integrate technical-metadata-service
For integrating the new technical metadata service:
1. TechMd robot marks the workflow step as started.
2. TechMd robot invokes TechMd service via HTTP call and exits without updating workflow step status.
3. TechMd service queues a job.
4. When job is executed, it performs techmd generation, and then updates workflow step (as either succeeded or failed).

(In the future when message broker, could get rid of robot and direct communication with workflow service.)

justinlittman:
closes #575

## Why was this change made?
Goodbye JHOVE! Hello technical-metadata-service.

No.


spec/robots/accession/technical_metadata_spec.rb
```diff
@@ -8,134 +8,97 @@
...
+                                        }
+                                      ]
+                                    }
+                                  }]
                                 })
       end
 
       before do
-        allow(Dor).to receive(:find).and_return(dor_object)
+        #   allow(Dor).to receive(:find).and_return(dor_object)
```
jcoyne: cruft?

Issue #586, opened 2020-02-21
Update etd robots to use legacy_metadata api 


jcoyne:
This decouples the writing to fedora 3, which is 1/2 of #582 (the reading part will have to wait until hydra_etd can be rewritten)

ndushay:
From https://github.com/sul-dlss/common-accessioning/issues/582#issuecomment-589858935: 

> It looks like completely decoupling the etd robots will not be possible until we modify hydra_etd to use the deposit endpoint. The etd robots depend on some non-standard datastreams (properties for one) that don't map to cocina models.

jcoyne:
Fixed by #591

Issue #582, opened 2020-02-21
[ÜBER] Decouple ETD robots from Fedora 3


jcoyne:
There should be no calls to `ETD.find`
in these classes:
EtdSubmit::CheckMarc
EtdSubmit::OtherMetadata
EtdSubmit::SubmitMarc

jcoyne:
It looks like completely decoupling the etd robots will not be possible until we modify hydra_etd to use the deposit endpoint.  The etd robots depend on some non-standard datastreams (properties for one) that don't map to cocina models.

ndushay:
I *think* PR #591 does this, not sure of the diff between this ticket and #586.

PR #591 gets rid of ETD.find calls;  it does not get rid of all coupling to Fedora 3.

jcoyne:
@ndushay #586 is a smaller scope than this.  This ticket is an uber ticket.

mjgiarlo:
I have marked this ticket as an über.

jcoyne:
Need to rewrite ETDs first because ETDs use non-standard datastreams, which are not supported by cocina.  Maybe we pull the robots stuff into the etd codebase.

PR #591, opened 2020-02-25
etd robots more decoupled from fedora 3

ndushay:
## Why was this change made?

Moves us a little further towards decoupling ETD robots from Fedora 3.   From https://github.com/sul-dlss/common-accessioning/issues/582#issuecomment-589858935: 

>completely decoupling the etd robots will not be possible until we modify hydra_etd to use the deposit endpoint. The etd robots depend on some non-standard datastreams (properties for one) that don't map to cocina models."

I'm not sure if this was part of (or all of) #582 or #586 - @jcoyne?

NOTE:  I am not confident the test code did a great job of ensuring that my changes are correct so careful review of my changes is welcome.  Or merge it and let it break.  whatever.

Also note:  I'm more inclined to do follow on PRs if this is incomplete in some way.

n/a


.rubocop_todo.yml
```diff
@@ -32,7 +32,7 @@ Lint/UnderscorePrefixedVariableName:
 
 # Offense count: 22
 Metrics/AbcSize:
-  Max: 98
+  Max: 100
```
jcoyne: 😱

lib/robots/dor_repo/etd_submit/other_metadata.rb
```diff
@@ -11,51 +11,51 @@ def initialize(opts = {})
...
-          attrs = { mimeType: 'application/xml', dsLabel: label, content: metadata }
-          datastream = ActiveFedora::Datastream.new(etd.inner_object, ds_name, attrs)
-          datastream.controlGroup = 'M'
-          datastream.save
+        def create_legacy_metadata(druid, new_metadata)
+          object_client = Dor::Services::Client.object(druid)
+
+          # yes, legacy_update will create the metadata
+          object_client.metadata.legacy_update(
+            content: {
```
jcoyne: This is the name of the datastream. In this case the content metadata. This should vary per datastream.

ndushay: d'oh!

.rubocop_todo.yml
ndushay: yeah, i know.  :-P

ndushay:
thanks for catching the error!  I have fixed it and added a test.

lib/robots/dor_repo/etd_submit/check_marc.rb
```diff
@@ -16,7 +16,8 @@ def initialize(opts = {})
 
         #========= This method check symphony and looks for MARC record based on druid ==========#
         def perform(druid)
-          etd = Etd.find(druid)
+          object_client = Dor::Services::Client.object(druid)
+          etd = object_client.find
```
jcoyne: It looks to me like this value gets passed to `Etd::IdentityMetadataGenerator.generate`, which requires an ActiveFedora Etd object.

lib/robots/dor_repo/etd_submit/other_metadata.rb
```diff
@@ -11,51 +11,55 @@ def initialize(opts = {})
           super('etdSubmitWF', 'other-metadata', opts)
         end
 
-        # create metadata datastreams for the work item
+        # create metadata for the work item
         def perform(druid)
-          etd = Etd.find(druid)
+          object_client = Dor::Services::Client.object(druid)
+          etd = object_client.find
```
jcoyne: It looks to me like this value gets passed to the *MetadataGenerators, which depend on an ActiveFedora Etd.

lib/robots/dor_repo/etd_submit/other_metadata.rb
```diff
@@ -11,51 +11,55 @@ def initialize(opts = {})
...
-          attrs = { mimeType: 'application/xml', dsLabel: label, content: metadata }
-          datastream = ActiveFedora::Datastream.new(etd.inner_object, ds_name, attrs)
-          datastream.controlGroup = 'M'
-          datastream.save
+        def create_legacy_metadata(druid, md_sym, new_metadata)
+          object_client = Dor::Services::Client.object(druid)
+
+          # legacy_update will create the metadata
+          object_client.metadata.legacy_update(
+            md_sym => {
```
jcoyne: This looks good. one optimization here is that we can generate all of the different content and pass them in a single `legacy_update` request.  e.g.:

```ruby
object_client.metadata.legacy_update(
  content: {
    updated: Time.now,
    content: new_metadata
  },
  identity: {
    ...
  },
  version: {
    ...
  },
  rights: {
    ...
  })

See http://sul-dlss.github.io/dor-services-app/#operation/metadata#update_legacy_metadata
```

spec/robots/etd_submit/check_marc_spec.rb
```diff
@@ -46,6 +45,7 @@
       end
 
       before do
+        allow(object_client).to receive(:find).and_return(object)
```
jcoyne: Dor::Services::Client::Object.find doesn't return a ActiveFedora Etd.

lib/robots/dor_repo/etd_submit/submit_marc.rb
```diff
@@ -67,12 +67,13 @@ def output_file(druid, record)
         end
 
         def generate_marc(druid)
-          etd = Etd.find(druid)
+          object_client = Dor::Services::Client.object(druid)
+          etd = object_client.find
```
jcoyne: These two calls are not equivalent, they return different classes of object.

spec/robots/etd_submit/submit_marc_spec.rb
```diff
@@ -57,20 +57,20 @@ def cleanup
...
-      it "creates marc and saves it to a file in the day's working directory" do
-        skip
-        # marc_file = File.join(ROBOT_ROOT, 'tmp', Time.now.strftime('%Y%m%d'), "#{@mock_workitem.druid.gsub(':','_')}.marc")
-        #         FileUtils.rm(marc_file) if File.exists(marc_file)
-        #
-        #         @robot.process_item(@mock_workitem)
-        #         File.should be_file(marc_file)
-      end
-    end
+    # describe '#process_items' do
```
jcoyne: Just remove this?

lib/robots/dor_repo/etd_submit/check_marc.rb
ndushay: so this should be left as `Etd.find()`?  or switch to `Dor::Etd.find` or `Dor::Item.find` or whatever?

lib/robots/dor_repo/etd_submit/other_metadata.rb
ndushay: same question as above.

lib/robots/dor_repo/etd_submit/other_metadata.rb
ndushay: will update

spec/robots/etd_submit/submit_marc_spec.rb
ndushay: sure.

.rubocop_todo.yml
ndushay: back to 98!!!  I didn't do it!!!!

.rubocop_todo.yml
```diff
@@ -148,7 +149,7 @@ RSpec/ScatteredLet:
   Exclude:
     - 'spec/robots/assembly/jp2_create_spec.rb'
 
-# Offense count: 6
+# Offense count: 3
```
ndushay: look!  I made it better!  🤣

PR #593, opened 2020-02-25
add spec for Dor::Etd::IdentityMetadataGenerator

### LINKED ISSUE #589: need tests for ETD identityMetadataGenerator
https://github.com/sul-dlss/common-accessioning/pull/585/files shows a new class split out for Dor::Etd::IdentityMetadataGenerator.  

It needs tests.  (It probably didn't have tests before it was split out - but that's no excuse).

ndushay:
## Why was this change made?

Fixes #589 

We had no test for this class, which was created in PR #585

na


lib/dor/etd/identity_metadata_generator.rb
```diff
@@ -53,17 +57,16 @@ def generate
               unless old_identity_doc.nil?
                 catkey = old_identity_doc.at_xpath('//catkey')
                 catkey = old_identity_doc.at_xpath("//otherId[@name='catkey']") if catkey.nil?
-                xml.text(catkey.text)
+                xml.text(catkey.text) if catkey
```
ndushay: look!  this was a bug I squashed after my tests found it!

jcoyne: The  ETDs are always going to have a catkey, so maybe we don't want a conditional.

ndushay: I wonder if there are any extant ETDs that aren't past this step that either don't have a catkey, or have an empty one.   If both of these are errors, there might be a better way to surface such a problem beyond "no method "text" on nil object".

I defer to your judgement -- it's just that in general, our metadata is in all sorts of states as requirements have changed over time.  For ETDs, I imagine there aren't a lot of old ones stuck in the middle of their accessioning.

We do have SOME, tho.  Perhaps @andrewjbtw can weigh in.

image

ndushay:
@andrewjbtw - do we ever have
-  ETDs with existing IdentityMetadata 
  - that are going through the etd-submit robots 
    - where the existing identityMetadata has an empty catkey?

("no" would be a delightful answer)

andrewjbtw:
I'd love to be able to say "no" but instead the accurate answer is "I have no idea" at this point.

ndushay:
@jcoyne took out the added if clause, added a comment, adjusted specs as necessary.

PR #609, opened 2020-03-10
accessioningWF: only generate techMd for preserved files

### LINKED ISSUE #608: Generate techMD only for files that are preserved
Currently, we are generating techMD for all files staged during accessioning. This includes files that are shelved and not preserved. We should generate techMD only for files being preserved, as the purpose of the techMD is to track information about files being stored for the long-term.

Example: 

This is a normal (i.e. non-Google) book where the jp2s are generated during assemblyWF and not shelved, and techMD has been generated on all of these files:

https://argo-stage.stanford.edu/view/druid:bs959cg1458?beta=true

ndushay:
## Why was this change made?

Fixes #608

we only want to generate techMd for preserved files.

n/a

not yet.  :-)

### DIFF
--- a/spec/robots/accession/technical_metadata_spec.rb
+++ b/spec/robots/accession/technical_metadata_spec.rb
@@ -2,6 +2,8 @@
 
 require 'spec_helper'
 
+RSpec::Matchers.define_negated_matcher :excluding, :include
+
 RSpec.describe Robots::DorRepo::Accession::TechnicalMetadata do
   subject(:robot) { described_class.new }
 
@@ -20,6 +22,7 @@ RSpec.describe Robots::DorRepo::Accession::TechnicalMetadata do
 
     context 'when a DRO with files' do
       let(:workspace) { File.absolute_path('spec/fixtures/workspace') }
+      let(:expected_file) { "file://#{workspace}/dd/116/zh/0343/dd116zh0343/content/folder1PuSu/story1u.txt" }
 
       let(:object) do
         Cocina::Models::DRO.new(externalIdentifier: 'druid:dd116zh0343',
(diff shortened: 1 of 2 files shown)


spec/robots/accession/technical_metadata_spec.rb
```diff
@@ -88,18 +126,99 @@
...
+              ),
+              headers: {
+                'Content-Type' => 'application/json',
+                'Authorization' => 'Bearer rake-generate-token-me'
+              }
+            )
+            .to_return(status: 200, body: '', headers: {})
+        end
+
+        it 'file is NOT sent to techmd service' do
```
ndushay: per cocina-models which uses default of false.

spec/robots/accession/technical_metadata_spec.rb
```diff
@@ -88,18 +126,99 @@
...
```
ndushay: per cocina models, which defaults to Administrative.new for File and that in turn gets preserve=false.

spec/robots/accession/technical_metadata_spec.rb
```diff
@@ -88,18 +126,99 @@
...
 
-      context 'when the DRO has no files' do
-        let(:object) do
-          Cocina::Models::DRO.new(externalIdentifier: '123',
-                                  type: Cocina::Models::Vocab.object,
-                                  label: 'my repository object',
-                                  version: 1)
+      context 'when file has true for sdrPreserve' do
+        before do
+          stub_request(:post, 'https://dor-techmd-test.stanford.test/v1/technical-metadata')
```
jermnelson: Not a blocker but seeing all of these stub requests in this spec I wonder if we should be creating an integration test as well.

ndushay: definitely!  I'll make a ticket for that.

ndushay: ticket existed - i expanded the description a bit.  https://github.com/sul-dlss/infrastructure-integration-test/issues/24

ndushay:
Made tickets #610 and  and sul-dlss/cocina-models/issues/87 re: default behavior when file is missing explicit sdrPreserve value

PR #656, opened 2020-05-06
Remove *-complete steps from accession WF

jcoyne:
## Why was this change made?

We can now use 'noop' and have dor-services-app set these steps to complete. This follows what we're doing in the technical metadata step

Ref https://github.com/sul-dlss/argo/issues/1789
Ref https://github.com/sul-dlss/dor-services-app/pull/850
Ref https://github.com/sul-dlss/workflow-server-rails/pull/377

no

Yes. Was tested on stage.


spec/robots/accession/shelve_spec.rb
```diff
@@ -33,15 +29,8 @@
       end
 
       it 'sets the shelve-complete step to completed' do
```
ndushay, line 31: test description is now incorrect?

PR #657, opened 2020-05-06
allow run_robot to work

peetucket:
## Why was this change made?

same change as sul-dlss/preservation_robots#222

the `#work` method in the robot now requires a context to be passed in for reporting to honeybadger, so the run_robot script needs to ~~pass in an extra parameter in the method call to work~~ call the `#perform` method instead, which takes care of the optional parameter

see https://github.com/sul-dlss/lyber-core/blob/master/lib/lyber_core/robot.rb#L40

not needed

no

### DIFF
--- a/bin/run_robot
+++ b/bin/run_robot
@@ -40,5 +40,5 @@ druids = if opts[:file]
          end
 
 druids.each do |druid|
-  bot.work druid
+  bot.perform druid
 end


mjgiarlo:
Should context be a required arg? If so, is this the only spot we need to make this change? Asking in @jcoyne's direction since this was a change he made: https://github.com/sul-dlss/lyber-core/pull/54 (I don't believe this is related to HB.)

peetucket:
Yeah, a better option than this would be to make the `context` arg optional and nil by default...makes everything backwards compatible.  Or maybe we should call `#perform` instead of `#work` from run_robot?

mjgiarlo:
@peetucket :speech_balloon: 
> Yeah, a better option than this would be to make the `context` arg optional and nil by default...makes everything backwards compatible. Or maybe we should call `#perform` instead of `#work` from run_robot?

Open to all this, or whatever, really. Let's see what @jcoyne says. I'm sure he had something in mind when he put in that PR.

jcoyne:
I was unaware that `#work` was part of the public API. Is there a reason we wouldn't do `klass.perform(druid)`?

peetucket:
#work isn't a private method, though it may not be used elsewhere directly, so I agree a cleaner solution is just to use the #perform method in the run_robot script -- i'll update the PRs

PR #663, opened 2020-05-13
Updates dor-services-client to 6.3.1.

justinlittman:
## Why was this change made?
Update to latest dor-services-client.

No.

No.


ndushay:
this seems worth running at least one integration test on stage that waits for object to be accessioned.

PR #680, opened 2020-07-07
Switch to circleci

aaron-collier:
## Why was this change made?

Moving away from travis

BUILD

README


mjgiarlo:
@aaron-collier do you think we should take this forward or walk it back?

aaron-collier:
Let me see if I can wrap this up. If not quickly (like with modsulator) I'll close this.

ndushay (approved):

PR #683, opened 2020-07-14
remove unused jhove setting

ndushay:
## Why was this change made?

b/c cruft makes maintenance harder

we stopped using jhove when we switch to techmd-service

checked and no jhove setting in shared_configs.


ndushay:
soooOOOOoooo - this actually runs on travis.  Or used to.

ndushay:
I hereby swear this is the only occurrence of "jhove" in this whole code base.

PR #695, opened 2020-09-09
add a test for member service

peetucket:
## Why was this change made?

~~(continues to) fixes sul-dlss/argo#2243~~

~~Basically we still have an issue with identifying members of a collection.  Object type is in a multivalued solr field, so that when dor-services-client parses this result from dor-services app here: https://github.com/sul-dlss/dor-services-client/blob/master/lib/dor/services/client/members.rb it makes the `type` an array in the response, and not a string. Thus, determining if a member is an item or a collection needs to be done assuming this value is an array, else you will always get an empty set when asking for items or collections regardless of how many members are returned.  We also need to guard against a nil type, as this may occur in some solr documents.~~

Add a new test for a class that had no test before.

Also see sul-dlss/dor-services-app#1058 and sul-dlss/dor-services-app#1060

Test suite

### DIFF
--- a/spec/lib/dor/release/member_services_spec.rb
+++ b/spec/lib/dor/release/member_services_spec.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Dor::Release::MemberService do
+  let(:member_service) { described_class.new(druid: 'druid:oo000oo0001') }
+  let(:item_member) { Dor::Services::Client::Members::Member.new(externalIdentifier: 'druid:xx828xx3282', type: 'item') }
+  let(:collection_member) { Dor::Services::Client::Members::Member.new(externalIdentifier: 'druid:xx222xx3282', type: 'collection') }
+  let(:response) do
+    [
+      collection_member,
+      item_member,
+      Dor::Services::Client::Members::Member.new(externalIdentifier: 'druid:xx828xx4444', type: nil)
+    ]
+  end
+
+  before { allow(member_service).to receive(:members).and_return(response) }
+
+  describe '#sub_collections' do
+    it 'returns only the sub collections in the collection' do
+      expect(member_service.sub_collections).to eq([collection_member])
+    end
+  end
+
+  describe '#items' do
+    it 'returns only the items in the collection' do
+      expect(member_service.items).to eq([item_member])
+    end
+  end
+end


lib/dor/release/member_service.rb
```diff
@@ -21,7 +21,7 @@ def sub_collections
       attr_reader :druid
 
       def members_of_type(type)
-        members.filter { |member| member.type == type }
+        members.filter { |member| member.type&.include? type }
```
jcoyne, line 24: Type of a Cocina::Model object is always a scalar (not an array). See https://github.com/sul-dlss/cocina-models/blob/master/lib/cocina/models/dro.rb#L26 However, it's a URI, so you want to use `collection?` or `dro?` (see https://github.com/sul-dlss/cocina-models/blob/master/lib/cocina/models/checkable.rb#L11)

spec/lib/dor/release/member_services_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Dor::Release::MemberService do
+  let(:member_service) { described_class.new(druid: 'druid:oo000oo0001') }
+  let(:member) { Struct.new(:externalIdentifier, :type, keyword_init: true) }
+  let(:item_member) { member.new(externalIdentifier: 'druid:xx828xx3282', type: ['item']) }
```
jcoyne, line 8: how about `instance_double(Cocina::Models::DRO, ...)` here?

spec/lib/dor/release/member_services_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  let(:collection_member) { member.new(externalIdentifier: 'druid:xx222xx3282', type: ['collection']) }
```
jcoyne, line 9: and how about `instance_double(Cocina::Models::Collection, ...)` here?

jcoyne (changes requested):

spec/lib/dor/release/member_services_spec.rb
peetucket: good idea, changed

spec/lib/dor/release/member_services_spec.rb
peetucket: good idea, changed

lib/dor/release/member_service.rb
peetucket: Yeah, it's not -- its returning an array ... I confirmed this by hitting the actual endpoint.  It's basically because it just grabs what comes from solr and sticks in the response (and solr has the object type in a multi-valued field).  

So I guess we are out of sync with the openapi and need to either fix DSA to return a scaler and then revert this change, or adjust openapi to reflect the fact that is an array.

I guess the real question is does it make sense for an object to be of two types.  I don't think so,  yet the solr field we derive this from is indexed in this way.

jcoyne: If we just fix DSA, then we don't need this change at all.  No need to revert anything.

peetucket: Yeah sorry, that's what I meant - not do this change in the PR here (though adjusting and keeping the test would be good).

jcoyne: I've made https://github.com/sul-dlss/dor-services-app/pull/1060 to fix the DSA problem.

peetucket: undid this change.

jcoyne (approved):

PR #697, opened 2020-09-16
Decommission commmon-accessioning-prod-e

tallenaz:
## Why was this change made?

See https://github.com/sul-dlss/operations-tasks/issues/2352

N/A

DevOpsDocs PR is forthcoming


jcoyne (approved):

tallenaz:
See https://github.com/sul-dlss/DevOpsDocs/pull/464

PR #744, opened 2021-03-23
manual bundle update to get cocina-models to latest

jmartin-sul:


jcoyne (approved):

jmartin-sul:
`assembly-objectfile` upgrade was breaking the build.  re-did the update without updating that gem.  see #745, which supersedes this PR.

PR #761, opened 2021-05-07
Add an enable lograge

aaron-collier:
## Why was this change made?

Install lograge for better log searching in cloudwatch

N/A


config/boot.rb
```diff
@@ -22,6 +22,9 @@
   #
   config.use_env = true
 
+  # Enable lograge
+  config.lograge.enabled = true
```
jcoyne, line 26: This is configuration for the Settings gem. I don't know what lograge is.

jcoyne (changes requested):

jcoyne (changes requested):
I don't think lograge is going to do anything for this project, since it doesn't use Rails.

aaron-collier:
@jcoyne cool. I'll close.

PR #770, opened 2021-06-18
move end-accession cleanup call to async

peetucket:
## Why was this change made?

Part of sul-dlss/argo#2752

Experiment in switching the end-accession/cleanup to a job. This allows the robot to add a lane-id and pass it along to dor-services-app.  Depends on a new version of dor-services-client: sul-dlss/dor-services-client#221

We also need to changes to dor-services-app and dor-services-client. Hold for these changes since they all need to be released together.

### DIFF
--- a/lib/robots/dor_repo/accession/end_accession.rb
+++ b/lib/robots/dor_repo/accession/end_accession.rb
@@ -18,7 +18,9 @@ module Robots
 
           # Call cleanup
           # Note that this used to be handled by the disseminationWF, which is no longer used.
-          Dor::Services::Client.object(druid).workspace.cleanup
+          # This is an asynchronous result. It will set the end-accession workflow step to complete when it is done.
+          object_client.workspace.cleanup(lane_id: lane_id(druid))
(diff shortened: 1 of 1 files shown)


ndushay:
@peetucket what is the goal of this?  Do you still believe it could improve a production situation, or is it now moot?

ndushay:
I see this is still part of a number of steps suggested.

peetucket:
The goal is to address https://github.com/sul-dlss/argo/issues/2752.  There is a lot of archeology in there trying to diagnose the intermittent failure, which leads to doing cleanup async (which is what this is starting to do).

ndushay:
I'm going to consider this "in progress" and put it in that column for SDR maint work cycle.  @peetucket if you want to fix the build and do some in situ testing, that would be great.

andrewjbtw:
I don't think https://github.com/sul-dlss/argo/issues/2752 has been fixed.

andrewjbtw:
Or will moving to sidekiq ultimately address this?

ndushay:
#2752 remains open, and this PR will stay in github, even when closed.  The open ticket has further discussion about other ways to do this ... so I'll leave this PR as closed.

PR #773, opened 2021-06-23
pass reading order to gem when generating content metadata from stub

### LINKED ISSUE #772: Keep book reading order the same throughout accessioning
**Describe the bug**
This item was registered as a book with a right-to-left reading order but after accessioning the reading order had been flipped to left-to-right: https://argo.stanford.edu/view/mr947cp4340 (I've now fixed it, so it's currently right-to-left).

In looking at the object history, I believe the reading order was flipped during the assemblyWF. The [contentMetadata history](https://sul-dor-prod.stanford.edu/fedora/objects/druid:mr947cp4340/datastreams/contentMetadata/history
) shows:

v0: `<bookData readingOrder="rtl"/>` (when registered, no files yet)
v1: `<bookData readingOrder="ltr"/>` (created through accessioning)
v2: `<bookData readingOrder="rtl"/>` (after I flipped the reading order back to right-to-left)

This object was accessioned via Goobi. I checked whether that could be the source of the issue:

- from talking with the digitization lab, this item was in the correct right-to-left reading order in the Goobi app
- Goobi creates a metadata file called "stubContentMetadata.xml" that is sent over with the content files and is used to create contentMetadata during assemblyWF
- the stubContentMetadata.xml for this object is in preservation core and has the correct reading order: `<content type="Book (rtl)">`

This leads me to believe the left-to-right reading order was introduced when the contentMetadata was created during assemblyWF.

**User Impact**
An extra step has to be taken to flip the reading order back to the right direction after accessioning. There are some large projects underway to digitize and accession right-to-left books, so this will slow those projects down by adding an extra step.

**To Reproduce**
I think this is going to happen to all right-to-left books being accessioned via Goobi. A second example with the same issue was accessioned yesterday: https://argo.stanford.edu/view/druid:ff217sb9425

**Expected behavior**
The reading order for books should not change after registration unless a user specifically changes it from the content type menu.

**Additional context**
We added the ability to handle reading order in Argo and Preassembly during the last workcycle. But we did not anticipate this issue, which seems specific to the accessioning path that uses stubContentMetadata.xml, something I think only Goobi does.

peetucket:
## Why was this change made?

Fixes #772 

~~[HOLD] Don't merge yet, pending some verification with actual data on stage~~  Done.

When books are accessioned via Goobi, the arrive with "stub" content metadata, which needs to be converted to content metadata during accessioning.  They will have the reading order identified in the content type, and this must be parsed correctly so we can send the reading order to our gem when generating the actual content metadata from the stub.  We recently added this new "reading_order" parameter to the content metadata generation gem, but do not currently set it when converting from stub --> content metadata during accessioning.  This fixes it.

Added tests.  Verified that it works correctly with actual stub content metadata coming from Goobi.

### DIFF
--- a/lib/dor/assembly/stub_content_metadata_parser.rb
+++ b/lib/dor/assembly/stub_content_metadata_parser.rb
@@ -26,7 +26,7 @@ module Dor
           end
         end
 
-        xml = ::Assembly::ContentMetadata.create_content_metadata(druid: @druid.druid, style: gem_content_metadata_style, objects: cm_resources, bundle: :prebundled, add_file_attributes: true)
+        xml = ::Assembly::ContentMetadata.create_content_metadata(druid: @druid.druid, style: gem_content_metadata_style, objects: cm_resources, bundle: :prebundled, add_file_attributes: true, reading_order: book_reading_order)
         @cm = Nokogiri.XML(xml)
         xml
       end
@@ -61,6 +61,13 @@ module Dor
         else
           :file # the default content metadata style if not found via the mapping is :file
         end
+      end
+
+      # this determines the reading order from the stub_object_type (default ltr unless the object type contains one of the possible "rtl" designations)
+      def book_reading_order
+        return 'rtl' if stub_object_type.include?('rtl') || stub_object_type.include?('r-l')
+
+        'ltr'
       end
 
       def stub_object_type
--- a/spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
+++ b/spec/lib/dor/assembly/stub_content_metadata_parser_spec.rb
@@ -5,7 +5,9 @@ require 'spec_helper'
 RSpec.describe Dor::Assembly::StubContentMetadataParser do
   describe '#stub_content_metadata_parser' do
     let(:druid) { 'druid:aa111bb3333' }
+    let(:druid_rtl) { 'druid:aa111bb5555' }
     let(:item) { Dor::Assembly::Item.new(druid: druid) }
+    let(:item_rtl) { Dor::Assembly::Item.new(druid: druid_rtl) }
 
     let(:xml_doc) do
       <<-EOXML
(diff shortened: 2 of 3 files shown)


peetucket:
Verified that actual original stub content metadata from https://argo.stanford.edu/view/mr947cp4340 will produce the correct reading order in contentMetadata using this code.  Next step is to deploy and verify with a new object.

peetucket:
@andrewjbtw I am fairly certain this will work, but do we want me to deploy to -stage so you can run a test there from Goobi?  The only project we need to deploy to see the change is common-accessioning

andrewjbtw:
I'm not sure we can test goobi in stage without some goobi-specific work to set up a process there. I'd probably need someone's help for that.

But I could make a test object with my own stubContentMetadata.xml that's copied from a Goobi one as a template. I can just drop that in /dor/assembly and run the workflow manually. Do you think that would work?

ndushay (approved):

peetucket:
> I'm not sure we can test goobi in stage without some goobi-specific work to set up a process there. I'd probably need someone's help for that.
> 
> But I could make a test object with my own stubContentMetadata.xml that's copied from a Goobi one as a template. I can just drop that in /dor/assembly and run the workflow manually. Do you think that would work?

I think that's a fine test.  This code change only affects the `content-metadata-create` step in assemblyWF, so if we get that step to run with the stubContentMetadata that didn't work before and see it keep the reading order as expected, it should be good to go.

ndushay:
just a nudge that maybe this could move forward again?

andrewjbtw:
Thanks for bringing that back up. I think I just need to test it on stage.

jcoyne:
@peetucket please rebase this first or it will cause errors on stage because the update-doi robot isn't in this branch.

peetucket:
Rebased on main.  Should we deploy the branch to -stage so Andrew can test?

jcoyne:
@peetucket go ahead

peetucket:
Ok, @andrewjbtw , this has been deployed to stage for further testing.  You should be able to do some goobi book tests now.

andrewjbtw:
Tested on stage and it looks good. There may be an issue with jp2-create in the stage environment, but the reading order in the contentMetadata came through correctly.

peetucket:
Ok, so will unhold this PR then.  Thanks

jcoyne (approved):

Issue #758, opened 2021-04-29
[monitor for problems] decrease robot worker count


jmartin-sul:
As discussed earlier this week, in an attempt to alleviate the [sul-dor-prod connection issues](https://github.com/sul-dlss/operations-tasks/issues/2432), we'll try decreasing the number of common-accessioning workers.  I think we'll want to keep an eye out for a few things:
1. robots being unable to keep up with overall accessioning demand due to reduced capacity (bad)
2. robots being unable to move things through at an acceptable rate when accessioning demand is heavier due to reduced concurrency (also bad)
3. a decrease in or elimination of Fedora connection errors as described in the above linked ops issue (good!) (the typical error is running out of outbound connections on dor-services-prod when trying to connect to sul-dor-prod, even after increasing the number of available outbound connections, e.g. [`Errno::EADDRNOTAVAIL: Failed to open TCP connection to sul-dor-prod.stanford.edu:443 (Cannot assign requested address - connect(2) for "sul-dor-prod.stanford.edu" port 443)`](https://app.honeybadger.io/projects/50568/faults/68024987))

The hypothesis is that at peak accessioning times, Fedora's unable to fulfill the demand that common-accessioning robots are placing on it (via dor-services-app) without backing up requests for long enough to exhaust DSA's available outbound connections.  The hope is that we can reduce the peak pressure on Fedora to keep from running out of outbound connections (while also keeping up with overall accessioning demand).  Circumstantial evidence in favor of this approach is that the problem most often coincides with large Argo bulk jobs involving opening/closing SDR objects.

The consensus was that we should decommission one of the four remaining common-accessioning VMs, as we did in #690 last fall.  However, as I went to look up those tickets to see what was done last time, I thought maybe it'd be better to start by decreasing the number of workers in the resque-pool config?  It seems like it'd have roughly the same effect (decreasing the common-accessioning concurrency and thus decreasing the max load common-accessioning can apply to Fedora at once).  But it'd be easier to revert if need be, and would keep us from needing to involve ops.  As another in between possibility, I think we could comment `common-accessioning-prod-d` out of the prod deploy config, kill resque-pool master on `common-accessioning-prod-d`, and then essentially have one of the common-accessioning boxes out of rotation till we add it back to the deployment config (also easier to revert by a single dev than decommissoning).

For now I'll put up a simple PR for decreasing the worker count, but if I'm missing something and we really should just decommission one, I'm happy to file an ops ticket and do the deploy config changes.

[...]

PR #774, opened 2021-06-26
reduce number of accessioning workers to alleviate pressure on fedora

jmartin-sul:
## Why was this change made?

we hope that this might reduce or eliminate issues with dor-services-app running out of outbound connections when talking to fedora (the hypothesis being the fedora is slow, and a backlog of requests to it will pile up as load from accessioning and/or indexing increases).

see https://github.com/sul-dlss/common-accessioning/issues/758 and https://github.com/sul-dlss/common-accessioning/pull/759/files

um...

worker allocation in resque pool config


ndushay (approved):

ndushay:
approved.  I'm always having to look all over to determine which worker counts are in shared_configs and which are in the apps themselves.

PR #809, opened 2021-12-13
Shut down common-accessioning workers

tallenaz:
## Why was this change made?

So that no work gets picked up by techmd while we migrate it to Ubuntu

Tested in stage

N/A


tallenaz:
Thanks, I can merge this when I move forward with the migration, either this afternoon or tomorrow.

PR #832, opened 2022-02-04
[HOLD] Transform contentMetadata to cocina on load

### LINKED ISSUE #827: accessioningWF:content step should load changes as cocina
Parse the contentMetadata.xml and transform the changes to cocina structural. This transform can be done by copying code over from DSA. Then stop using the legacy deposit endpoint and use the cocina update endpoint.

jcoyne:
Hold until @andrewjbtw has tested this.

## Why was this change made?

Fixes #827

### DIFF
--- a/lib/dor/file_sets.rb
+++ b/lib/dor/file_sets.rb
@@ -1,0 +1,94 @@
+# frozen_string_literal: true
+
+module Dor
+  # builds the FileSet instance from a Dor::Item
+  class FileSets
+    # @param [Nokogir::XML::Document] ng_xml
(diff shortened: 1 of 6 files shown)


mjgiarlo (approved):

mjgiarlo (approved):
This time with a comment: it looks like there's been a noticeable coverage drop on some lines we may care about covering. Reasonable to add a test or two? 🤷🏻‍♂️

jcoyne:
Andrew asks that this code should copy the rights from the object to all the newly added files.

PR #842, opened 2022-02-23
Ensure that dark objects do not try to publish or shelve

### LINKED ISSUE #841: Override preassembly/assembly-generated publish/shelve settings when object rights are dark
**Describe the issue**
Some objects with object-level rights set to dark are stalling in accessioning because their publish and shelve settings conflict with the rights setting of dark.

**Background**
When an object is set to dark, it should not have a Purl and any file content within that object should not be copied to Stacks. "Dark" rights essentially mean that an object should be stored in the preservation system only. At the file level, this corresponds to the following publish/shelve/preserve settings:

- publish: no
- shelve: no
- preserve: yes

We have a check that when object rights are dark, the publish and shelve permissions must be set to no. When an existing object has its rights changed to dark via Argo, this automatically sets the file rights to dark and changes the publish and shelve settings to "no". This has been working as expected.

**The problem**

Under some circumstances, the "publish" and "shelve" permissions on files can be set to "yes" even though the object rights set to dark:

- the assemblyWF may set these permissions automatically based on mimetype
- a Preassembly user may set these permissions in a file manifest

When this happens, items stall during accessioning with an error of the form:

```
content-metadata : Bad Request: 400 ({"errors":[{"status":"400","title":"Bad Request","detail":"Not all files have dark access and/or are unshelved when item access is dark: [\"jy889mc9982_sl.m4a\", \"jy889mc9982_img_1.jp2\", \"jy889mc9982_img_2.jp2\", \"jy889mc9982_img_3.jp2\", \"jy889mc9982_img_4.jp2\"]"}]})
```

This error occurs at a point in the accessionWF where users cannot correct it because the object is locked and the publish and shelve settings are not accessible via Argo. It might be possible to fix the issue by editing the contentMetadata.xml on the /dor filesystem but most people do not have that level of access, nor does it seem like a good idea to make a habit of editing those files.

**Proposal**

When object rights are set to dark, ensure that the publish and shelve settings are automatically set to "no". I'm not sure of the best way to do this. It could be done by:

- making sure Preassembly generates the publish and shelve settings by checking the rights first and flipping them to "no" even if the user had set them to yes in a file manifest
- setting the publish/shelve to "no" when running through the assemblyWF
- setting the publish/shelve to "no" when translating the contentMetadata.xml into Cocina structural metadata in the accessionWF

**Additional information**

Another option for the Preassembly case would be to have the job check the settings before starting accessioning and give the user a chance to change them. However, the UI doesn't have a clear way to give users a message about why a Preassembly job failed.

jcoyne:
## Why was this change made?

Fixes #841

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -20,6 +20,9 @@ Naming/PredicateName:
   ForbiddenPrefixes:
     - is_
 
+RSpec/NestedGroups:
+  Max: 5
+
 Gemspec/DateAssignment: # new in 1.10
   Enabled: true
 Gemspec/RequireMFA: # new in 1.23
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -74,10 +74,6 @@ RSpec/MultipleExpectations:
 RSpec/MultipleMemoizedHelpers:
   Max: 11
 
-# Offense count: 13
-RSpec/NestedGroups:
-  Max: 4
-
 # Offense count: 16
 # Configuration parameters: IgnoreNameless, IgnoreSymbolicNames.
 RSpec/VerifiedDoubles:
--- a/lib/dor/structural_metadata.rb
+++ b/lib/dor/structural_metadata.rb
@@ -15,7 +15,22 @@ module Dor
     end
 
     def update
-      (cocina_item.structural || Cocina::Models::DROStructural).new(props)
+      value = (cocina_item.structural || Cocina::Models::DROStructural).new(props)
+      overwrite_dark_access(value)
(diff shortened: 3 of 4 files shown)


.rubocop.yml
```diff
@@ -20,6 +20,9 @@ Naming/PredicateName:
   ForbiddenPrefixes:
     - is_
 
+RSpec/NestedGroups:
+  Max: 5
+
```
mjgiarlo, line 25: Yes, please.

justinlittman (approved):

PR #888, opened 2022-04-28
use cocina-models factories for specs

### LINKED ISSUE #887: Use cocina factory for spec
To reduce boilerplate.

See https://github.com/sul-dlss/dor-services-app/pull/3948

ndushay:
You might want to review the commits separately

to reduce boilerplate in our specs.

Fixes #887 

specs


spec/lib/dor/structural_metadata_spec.rb
```diff
@@ -51,21 +51,7 @@
 
     let(:apo_druid) { 'druid:pp000pp0000' }
```
justinlittman, line 52: `apo_druid` no longer necessary.

spec/robots/assembly/checksum_compute_spec.rb
```diff
@@ -4,7 +4,8 @@
 
 RSpec.describe Robots::DorRepo::Assembly::ChecksumCompute do
   let(:robot) { described_class.new }
-  let(:druid) { 'aa222cc3333' }
+  let(:druid) { 'bb222cc3333' }
```
justinlittman, line 7: For what it is worth, in other places we use `druid` and `bare druid`.

justinlittman (approved):
Approve pending consideration of comments.

Doesn't it feel good to throw code away?

spec/robots/assembly/checksum_compute_spec.rb
ndushay: true.  fixed.

PR #893, opened 2022-05-04
Map MODS to cocina before updating.

### LINKED ISSUE #891: Map MODS to cocina before updating
Instead of using the `update_mods` API call, use the functions in cocina-models to convert to cocina first.

```
Cocina::Models::Description.new(
  Cocina::Models::Mapping::FromMods::Description.props(mods: mods_ng, druid: cocina.externalIdentifier, label: cocina.label)
)
```

Do that here:
https://github.com/sul-dlss/common-accessioning/blob/5b75f0ca9a68f84e13c86a62625c67e45f0658fc/lib/robots/dor_repo/accession/descriptive_metadata.rb#L18-L19

This is similar to the work in: https://github.com/sul-dlss/argo/pull/3614

justinlittman:
closes #891

Allow retiring the mods update endpoint.

Unit


justinlittman:
P.S. I made additions to shared configs.

config/settings.yml
```diff
@@ -32,3 +32,5 @@ workflow:
 
 redis:
   url: ~
+
+purl_url: 'https://purl-example.stanford.edu'
```
jcoyne, line 36: Has this change been made in shared configs for the 3 environments?

lib/robots/dor_repo/base.rb
```diff
@@ -5,6 +5,11 @@ module DorRepo
     class Base
       include LyberCore::Robot
 
+      def initialize(...)
+        super
+        Cocina::Models::Mapping::Purl.base_url = Settings.purl_url
```
jcoyne, line 10: Can this be in the DescriptiveMetadata robot or in config/boot.rb ?

jcoyne (changes requested):

config/settings.yml
justinlittman: Yes.

lib/robots/dor_repo/base.rb
justinlittman: Ahh, I did not know about `boot.rb`. Fixed.

jcoyne (approved):

PR #920, opened 2022-06-15
Remove unnecessary conditionals

jcoyne:

This makes it easier to read the code.  We never set these to anything except 'false', so we can just hardcode the values.  See https://github.com/sul-dlss/shared_configs/blob/common-accessioning-prod/config/settings/production.yml#L11-L12

CI

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -8,8 +8,6 @@ assembly:
     - spec/test_input2
   cm_file_name: contentMetadata.xml # the name of the contentMetadata file
   stub_cm_file_name: stubContentMetadata.xml # the name of the stub contentMetadata file
-  overwrite_jp2: false # indicates if the jp2-create robot should overwrite an existing jp2 of the same name as the new one being created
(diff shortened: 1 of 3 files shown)


jcoyne:
@ndushay because I looked at the production configs and that's what we have set there.

andrewjbtw:
> PR looks fine. Question: how are you sure that we never set these to `true` ?
> 
> Also wondering if @andrewjbtw has any need for the overwriting functionality, even if it's not used now.

I'm very late to this, as it's now closed, but I think over time things settled on always using the same options for these two settings, after a period of trying them out different ways. The current settings make it possible for us to manually supply jp2s without them getting overwritten by the workflow step.

ndushay:
> > PR looks fine. Question: how are you sure that we never set these to `true` ?
> > Also wondering if @andrewjbtw has any need for the overwriting functionality, even if it's not used now.
> 
> I'm very late to this, as it's now closed, but I think over time things settled on always using the same options for these two settings, after a period of trying them out different ways. The current settings make it possible for us to manually supply jp2s without them getting overwritten by the workflow step.

Thanks for the reassurance!

PR #921, opened 2022-06-16
Notify about a path we think might be dead code

jcoyne:

We weren't seeing the messages from this path in the logs. Maybe this never happens?  If we don't see this notification in a few months, we can remove this code.

it wasn't! It's dead code 💀


ndushay (approved):
"How am I driving?  Call W-RIT-EAT-EST!"

PR #925, opened 2022-06-23
work with assembly-image without dpg processing methods

ndushay:

The DPG special folders way of accessioning is no longer used;  this PR removes the unneeded check for DPG folder convention.

specs and also Andrew is doing testing in stage environment.


spec/robots/assembly/jp2_create_spec.rb
```diff
@@ -497,8 +497,8 @@
                                                   administrative: { publish: false, sdrPreserve: false, shelve: false } },
                                                 { type: 'https://cocina.sul.stanford.edu/models/file',
                                                   externalIdentifier: 'https://cocina.sul.stanford.edu/file/661f8462-e0c2-49c4-b617-4d5c11a969b7',
-                                                  label: 'hh222cc3333_05_001.jp2',
-                                                  filename: 'hh222cc3333_05_001.jp2',
+                                                  label: 'hh222cc3333_00_001.jp2',
+                                                  filename: 'hh222cc3333_00_001.jp2',
```
ndushay, line 501: this test was failing without this change.  I think it's because of how the mocking was set up, but this was the easiest way to get it working.

lib/robots/dor_repo/assembly/jp2_create.rb
```diff
@@ -55,15 +55,10 @@ def create_jp2s(assembly_item, cocina_model)
...
-                        message = "WARNING: Did not create jp2 for #{assembly_image.path} -- since another JP2 with the same DPG base name called #{assembly_image.dpg_jp2_filename} exists"
-                        LyberCore::Log.warn(message)
-                        new_jp2_file_name(file_node, assembly_image.dpg_jp2_filename, assembly_image.path)
-                      elsif File.exist?(assembly_image.jp2_filename)
+          file_name = if File.exist?(assembly_image.jp2_filename)
                         # if we have an existing jp2 with the same basename as the tiff -- don't fail, but do log it
                         Honeybadger.notify('Does this path ever get hit in production?  If you see this error, ' \
-                                           "please write a test for this case in common-accessioning, presently it's not covered.", { image_path: img.path })
+                                           "please write a test for this case in common-accessioning, presently it's not covered.", { image_path: assembly_image.path })
                         message = "WARNING: Did not create jp2 for #{assembly_image.path} -- file already exists"
```
peetucket, line 62: I presume this `img.path` reference was broken before and we just never saw it

peetucket (approved):

lib/robots/dor_repo/assembly/jp2_create.rb
ndushay: yes, correct.

PR #930, opened 2022-06-28
add note about libvips to README

ndushay:

because common-accessioning now requires libvips.  (it used to require ImageMagick, but that was never documented.  Yay us.)


ndushay:
put this commit into PR #929 instead.

PR #945, opened 2022-07-07
Move the ContentMetadata code in from assembly-objectfile

jcoyne:

The only place that this code is used in DLSS is to convert stubContentMetadata.xml to contentMetadata.xml in this repo.  It's easier to manage if the code is where it is used.

Ci & integration tests


ndushay (approved):
great idea!  though it does make me ponder if building cocina structural would be usefully done in assembly-objectfile.  I think I've pondered that before and always landed on "no."

jcoyne:
> great idea! though it does make me ponder if building cocina structural would be usefully done in assembly-objectfile. I think I've pondered that before and always landed on "no."

Another alternative we could do here is to convert stubContentMetadata.xml right to the cocina API and skip the contentMetadata.xml step. That would allow us to throw all this code out.

ndushay:
Marking "Hold" for integration tests to be run.

jcoyne:
@ndushay I'm happy to run the integration tests, however we don't have an integration test for stubContentMetadata (the only path that uses this new code), so these paths won't be covered.

jcoyne:
Integration tests have successfully run.

PR #958, opened 2022-07-11
Use real files for testing

jcoyne:

Now that accessioning-objectfile raises an error when the files are bad, we need to have real files in place for testing

CI


ndushay (approved):
merge when green

PR #962, opened 2022-07-13
[refactor] Get rid of unused 'relative_path' option

jcoyne:
We never use this option, so we can simplify the code


ndushay (approved):
approving;  needs a rebase.

PR #966, opened 2022-07-13
[refactor] Remove FileSetBuilder

jcoyne:

This class is now so simple, we can replace it with a single line


ndushay (approved):

ndushay:
(I guess I mean pre-assembly)

PR #968, opened 2022-07-13
[refactor] Replace style with object type

jcoyne:

The style obfuscates what is happening and has no relevance in this repo

CI

### DIFF
--- a/lib/dor/assembly/content_metadata_from_stub.rb
+++ b/lib/dor/assembly/content_metadata_from_stub.rb
@@ -2,8 +2,6 @@
 
 module Dor
   module Assembly
-    VALID_STYLES = %i[simple_image simple_book file map 3d].freeze
-
     # This class generates content metadata XML from stubContentMetadata
     class ContentMetadataFromStub
       # Generates image content XML metadata for a repository object.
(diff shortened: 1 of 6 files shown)


lib/dor/assembly/content_metadata.rb
```diff
@@ -49,19 +44,6 @@ def self.find_common_path(objects)
         ::Assembly::ObjectFile.common_path(all_paths) # find common paths to all files provided if needed
       end
       private_class_method :find_common_path
-
-      def self.object_level_type(style)
-        raise "Supplied style (#{style}) not valid" unless VALID_STYLES.include? style
```
ndushay, line 54: doesn't this PR lose this validity check?

lib/dor/assembly/content_metadata/file_set.rb
```diff
@@ -10,14 +10,14 @@ class FileSet
...
         # @param [Array<Assembly::ObjectFile>] resource_files
-        def initialize(resource_files:, style:)
+        def initialize(resource_files:, object_type:)
           @resource_files = resource_files
-          @style = style
+          @object_type = object_type
         end
 
-        # otherwise look at the style to determine the resource_type_description
+        # otherwise look at the object_type to determine the resource_type_description
```
ndushay, line 20: "otherwise" is an odd way to start a method comment  - can you please improve on it?

lib/dor/assembly/content_metadata.rb
```diff
@@ -14,29 +12,26 @@ class ContentMetadata
...
-      #                 :simple_image (default), contentMetadata type="image", resource type="image"
-      #                 :file, contentMetadata type="file", resource type="file"
-      #                 :simple_book, contentMetadata type="book", resource type="page", but any resource which has file(s) other than an image, and also contains no images at all, will be resource type="object"
-      #                 :book_with_pdf, contentMetadata type="book", resource type="page", but any resource which has any file(s) other than an image will be resource type="object" - NOTE: THIS IS DEPRECATED
-      #                 :book_as_image, as simple_book, but with contentMetadata type="book", resource type="image" (same rule applies for resources with non images)  - NOTE: THIS IS DEPRECATED
-      #                 :map, like simple_image, but with contentMetadata type="map", resource type="image"
-      #                 :3d, contentMetadata type="3d", ".obj" and other configured 3d extension files go into resource_type="3d", everything else into resource_type="file"
-      #                 :webarchive-seed, contentMetadata type="webarchive-seed", resource type="image"
```
ndushay, line 25: this code is still used for :webarchive-seed, yes?  probably crawls also

ndushay (changes requested):
I love this clarification.  I have a few points of confusion:

- does the code need to do anything for webarchive object types?  

I would be more comfortable if this change was accompanied by running integration tests that accession objects needing contentMetadata.

lib/dor/assembly/content_metadata.rb
jcoyne: Yes, but the check is unnecessary as the inputs are limited by what is provided in this repo.

lib/dor/assembly/content_metadata.rb
jcoyne: No, this is only for stubMetadata stuff from Goobi.

jcoyne:
@ndushay
> I love this clarification. I have a few points of confusion:
>
>   does the code need to do anything for webarchive object types?
>
> I would be more comfortable if this change was accompanied by running integration tests that accession objects needing contentMetadata.

This is only for Goobi objects, for which we don't have any integration tests AFAIK.  Webarchive objects do not use this code.

jcoyne:
Since there's some confusion about what code path this is for, I made https://github.com/sul-dlss/common-accessioning/pull/970 to clarify.

ndushay:
I've just made a ticket for an integration test for an object coming from goobisul-dlss/infrastructure-integration-test/issues/459) , and I would suggest that putting that together would create an appropriate safety net for this work.

lib/dor/assembly/content_metadata/file_set.rb
jcoyne: I'm just going to delete this comment. I'm not sure what it's trying to say.

jcoyne:
@ndushay as far as I know, we don't have a stage environment for Goobi. Without that, just the unit tests that have things we think looks like stubContentMetadata.xml is just as good, right?

ndushay (approved):

ndushay (dismissed):
waiting for result of discussion of goobi integration test.

jcoyne:
Integration tests pass.

peetucket (approved):

PR #969, opened 2022-07-13
Convert stubContentMetadata directly to cocina structural

jcoyne:

Don't bother creating contentMetadata.xml first.  In the future there will be no need for contentMetadata.xml, only stubContentMetadata (from goobi) will be used.

CI & new goobi integration test.


lib/dor/assembly/stub_content_metadata_parser.rb
```diff
@@ -45,26 +43,11 @@ def load_stub_content_metadata
...
-          'image'
-        else
-          'file' # the default content metadata style if not found via the mapping
-        end
-      end
-
       # this determines the reading order from the stub_object_type (default ltr unless the object type contains one of the possible "rtl" designations)
       def book_reading_order
-        return 'rtl' if stub_object_type.include?('rtl') || stub_object_type.include?('r-l')
+        return 'right-to-left' if stub_object_type.include?('rtl') || stub_object_type.include?('r-l')
```
aaron-collier, line 48: can we safely change this or do we need to make sure access is ok with it?

aaron-collier (approved):

PR #970, opened 2022-07-13
[refactor] Move ContentMetadata to StubContentMetadata

jcoyne:
To provide clarity on what this module does


ndushay (changes requested):
I'm not sure I agree with the naming.  In more words some of these (all of these?) classes are about creating ContentMetadata _from_ StubContentMetadata.    I wouldn't expect them all to be called "stubContentMetadata".   Don't we more often name things for what they become, rather than what they started as?

I would instead welcome comments explaining that the contentMetadata is being generated from stub contentMetadata .. and that stub contentMetadata comes from goobi (I didn't know that until recently.)

jcoyne:
@ndushay okay, I renamed to ContentMetadataFromStub

ndushay (approved):

PR #997, opened 2022-09-15
Consider object access when determining default file rights in assemblyWF

### LINKED ISSUE #996: assemblyWF should handle dark rights appropriately
**Describe the bug**
The assemblyWF is hitting an error when people try to accession druids with dark rights. The rights to the files should automatically be set to dark in this context. Instead, it appears that the rights are being set to something else - or possibly are not set at all? - which is resulting in a validation error:

```
content-metadata-create : Not all files have dark access and/or are unshelved when object access is dark: ["nj233pb3063_Lozano_Mayra_Page_01.pdf", "nj233pb3063_Lozano_Mayra_Page_01.xml", 
```

**User Impact**
This is blocking accessioning of items when rights are set to dark. The specific cases where this is a problem currently are all accessions from Goobi. I have not yet checked if Preassembly has this issue.

See these three items for examples:
https://argo.stanford.edu/catalog?f%5Bwf_wps_ssim%5D%5B%5D=assemblyWF%3Acontent-metadata-create%3Aerror

**Expected behavior**
When content is registered with dark rights, the assemblyWF should process them with the correct settings.

mjgiarlo:

Fixes #996

This commit effectively coerces shelve to false at the file level when the object is dark.

CI. @andrewjbtw chatted, and after I ran the integration tests in stage, he gave the OK to merge and deploy to production.


justinlittman (approved):

andrewjbtw:
Looks like this only changes shelve to false. Publish also has to be false in this situation or you'll end up with broken links since the files will not be on stacks.

PR #998, opened 2022-09-15
[HOLD] Include presentation metadata only for select image types

mjgiarlo:

Fixes sul-dlss/assembly-objectfile#102

This commit further restricts which images receive presentation information (height/width) gathered by exif to the types currently returned by the existing `#valid_image?` method in the public API. These include TIFF, PNG, JPEG, and JP2. This prevents recording (perhaps bogus or nil) presentation information for image types that are unsupported for presentation elsewhere in SDR (e.g., the sul-embed service).

CI and will test in a deployed env with @andrewjbtw

### DIFF
--- a/lib/robots/dor_repo/assembly/exif_collect.rb
+++ b/lib/robots/dor_repo/assembly/exif_collect.rb
@@ -32,7 +32,10 @@ module Robots
               file[:size] = object_file.filesize if !file[:size] || file[:size].zero?
               file[:hasMimeType] ||= object_file.mimetype
 
-              next unless object_file.image?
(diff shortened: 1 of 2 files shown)


jcoyne (changes requested):
Should this be "Fixes https://github.com/sul-dlss/assembly-objectfile/issues/102"?

mjgiarlo:
@jcoyne 💬 
> Should this be "Fixes [sul-dlss/assembly-objectfile#102](https://github.com/sul-dlss/assembly-objectfile/issues/102)"?

Yep! Good catch. Fixed and superseded by #999. (The typo was in the commit message and also the branch name.)

PR #1000, opened 2022-09-15
Turn off publish and shelve when dark in missed spots

mjgiarlo:

This commit patches more code paths that set the publish and shelve attributes, ensuring that dark objects have *no* files set to publish or shelve EVEN IF the stub content metadata says so. This is a problem with some Goobi objects, and this patch allows those items to continue accessioning.

CI, and will test in a deployed env with @andrewjbtw

### DIFF
--- a/lib/dor/assembly/content_metadata_from_stub/structural_builder.rb
+++ b/lib/dor/assembly/content_metadata_from_stub/structural_builder.rb
@@ -34,11 +34,11 @@ module Dor
         end
         private_class_method :find_common_path
 
-        def self.administrative(file_attributes)
+        def self.administrative(file_attributes, object_access)
           {
             sdrPreserve: file_attributes[:preserve] == 'yes',
-            publish: file_attributes[:publish] == 'yes',
-            shelve: file_attributes[:shelve] == 'yes'
+            publish: object_access.view == 'dark' ? false : file_attributes[:publish] == 'yes',
+            shelve: object_access.view == 'dark' ? false : file_attributes[:shelve] == 'yes'
           }
         end
         private_class_method :administrative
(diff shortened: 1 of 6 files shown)


lib/dor/assembly/content_metadata_from_stub/structural_builder.rb
```diff
@@ -34,11 +34,11 @@ def self.find_common_path(objects)
...
 
-        def self.administrative(file_attributes)
+        def self.administrative(file_attributes, object_access)
           {
             sdrPreserve: file_attributes[:preserve] == 'yes',
-            publish: file_attributes[:publish] == 'yes',
-            shelve: file_attributes[:shelve] == 'yes'
+            publish: object_access.view == 'dark' ? false : file_attributes[:publish] == 'yes',
+            shelve: object_access.view == 'dark' ? false : file_attributes[:shelve] == 'yes'
           }
```
mjgiarlo, line 42: The logic beforehand deferred entirely to the stub content metadata when determining if a file should be set to publish or shelve. We now consider the object rights, and if dark, coerce to `false` for both.

lib/dor/file_sets.rb
```diff
@@ -120,7 +120,7 @@ def build_files(file_nodes)
     end
 
     def file_administrative(node)
-      default_administrative = self.class.default_administrative_attributes(node['mimetype'])
+      default_administrative = self.class.default_administrative_attributes(node['mimetype'], object_access: dro_access)
```
mjgiarlo, line 123: This little bit should have been part of the prior PR, but I missed it.

spec/fixtures/workspace/ab/123/cd/4567/content_metadata.xml
```diff
@@ -4,14 +4,14 @@
       <attr name="pageNumber">3</attr>
       <attr name="pageLabel">ii</attr>
       <attr name="googlePageTag">IMAGE_ON_PAGE,IMPLICIT_PAGE_NUMBER</attr>
-      <file id="00000001.jp2" format="JPEG2000" mimetype="image/jp2" size="169627" shelve="no" deliver="no" preserve="yes">
+      <file id="00000001.jp2" format="JPEG2000" mimetype="image/jp2" size="169627" shelve="no" publish="yes" preserve="yes">
```
mjgiarlo, line 7: Changes to this file were made to update what the content metadata should look like.

peetucket (approved):

PR #1014, opened 2022-10-27
Adds basepath to techmd call.

justinlittman:
refs https://github.com/sul-dlss/technical-metadata-service/issues/398

To call techmd service with the now required basepath param.

Unit

### DIFF
--- a/lib/preserved_file_uris.rb
+++ b/lib/preserved_file_uris.rb
@@ -22,13 +22,16 @@ class PreservedFileUris
     @filepaths ||= filepath_uris.map(&:filepath)
   end
 
+  def content_dir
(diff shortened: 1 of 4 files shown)


lib/preserved_file_uris.rb
```diff
@@ -22,13 +22,16 @@ def filepaths
...
   private
 
   attr_reader :obj, :druid
 
-  def content_dir
-    workspace = DruidTools::Druid.new(druid, File.absolute_path(Settings.sdr.local_workspace_root))
-    workspace.content_dir(false)
+  def local_workspace_root
+    @local_workspace_root ||= File.absolute_path(Settings.sdr.local_workspace_root)
   end
```
ndushay, line 35: this is a straight refactor, yes?

ndushay (approved):

PR #1047, opened 2023-03-27
Update to Sidekiq7

justinlittman:
7 > 6


Gemfile.lock
```diff
@@ -192,10 +192,11 @@ GEM
       pry (>= 0.13, < 0.15)
     public_suffix (5.0.1)
     racc (1.6.2)
-    rack (2.2.6.4)
+    rack (3.0.7)
```
edsu, line 195: Are there any concerns about moving a major version of rack too? Are we on v2 most other places?

edsu (approved):

Gemfile.lock
justinlittman: There's no web app, so shouldn't be a problem.

Issue #1043, opened 2023-03-10
goobi-notify step in goobiWF is hitting errors more frequently for no apparent reason


andrewjbtw:
**Describe the bug**
When Goobi users register items in Argo, they set the items to use the goobiWF. This workflow takes some of the SDR metadata (druid, tags, etc.) and transmits it to Goobi. This makes it possible to associate SDR druids with files being processed in Goobi.

The last step in this workflow is `goobi-notify`. This step has been hitting more errors in recent weeks than it did before. The puzzling thing about it is that there's no clear indication that something is wrong.

In the past, when this error occurred there was usually a clear reason for that, such as a missing tag. Goobi expects certain data to come over from SDR and if that data is not received, it reports an error. When that happens, users can add the missing information in Argo and then retry the step.

In the recent errors, there hasn't been missing information. When users have retried the step, it's worked right away with no further changes. We have not been able to determine what is going wrong.

The goobi error message is pretty generic:
```
Error: goobi-notify : Bad Request: 400 (<?xml version="1.0" encoding="UTF-8" standalone="yes"?><creationResponse><errorText>Process template Example_Workflow does not exist.</errorText><processId>0</processId><processName>sul:M1132_s2_b38_f9</processName><result>error</result></creationResponse>) 
```
The error text suggests that a template couldn't be found. It might be the case that SDR is sending incomplete information to Goobi the first time and then sending the correct information after a retry.

**User Impact**
It disrupts Goobi users work. They have to dig around in Argo to figure out why their items are not appearing in Goobi. There doesn't seem to be a pattern of when it will fail.

**To Reproduce**
I attempted to reproduce this issue on stage but haven't been successful. I registered 100 items to use Goobi in a relatively big batch on the theory that larger batches might be more likely to generate errors.

This item currently has the error and in the short term it's being left alone to provide an example:
https://argo.stanford.edu/view/druid:gt114mr9927

**Expected behavior**
If all required information is provided during registration, goobi-notify should succeed.

jmartin-sul:
see e.g. https://app.honeybadger.io/projects/52894/faults/94132373

peetucket:
Made an attempt here: https://github.com/sul-dlss/dor-services-app/pull/4467 but didn't seem to make a difference.  No other ideas at the moment.

mjgiarlo:
Closed by https://github.com/sul-dlss/argo/pull/4018

PR #1053, opened 2023-04-26
Retry goobi-notify immediately if we get an apparently bogus error

mjgiarlo:

Connects to #1043

This commit allows the goobi-notify robot to retry automatically when DSC raises a bad request error with a particular body.

CI

### DIFF
--- a/lib/robots/dor_repo/goobi/goobi_notify.rb
+++ b/lib/robots/dor_repo/goobi/goobi_notify.rb
@@ -15,6 +15,9 @@ module Robots
         # @param [String] druid -- the Druid identifier for the object to process
         def perform_work
           object_client.notify_goobi
+        rescue Dor::Services::Client::BadRequestError => e
+          # NOTE: We don't know why this error is occurring, but the condition seems to go away momentarily, so retry automatically.
+          retry if e.message.match?(/Process template Example_Workflow does not exist/)
         end
       end
     end


lib/robots/dor_repo/goobi/goobi_notify.rb
```diff
@@ -15,6 +15,9 @@ def initialize
         # @param [String] druid -- the Druid identifier for the object to process
         def perform_work
           object_client.notify_goobi
+        rescue Dor::Services::Client::BadRequestError => e
+          # NOTE: We don't know why this error is occurring, but the condition seems to go away momentarily, so retry automatically.
+          retry if e.message.match?(/Process template Example_Workflow does not exist/)
         end
```
peetucket, line 21: While this is an interesting idea,  if this error occurs, the robot has no idea why (missing the process tag?  typo in the process tag?  totally bogus and can be quickly retried?)  For the first two cases, a retry will not help, and while we can detect the "missing a tag' case, but not the other case.  This would lead to an infinite retry loop.

mjgiarlo: Yes, it would.

ndushay:
Noting here that this was deemed a bad idea in a slack thread as legit errors could be retried forever. https://stanfordlib.slack.com/archives/C09M7P91R/p1683026791898829

PR #1062, opened 2023-05-22
[HOLD] verify checksums exist

### LINKED ISSUE #1056: Verify required checksums before finishing checksum-compute (assemblyWF)
**Describe the bug**
In the assemblyWF, there is a step called `checksum-compute`. I think this step does at least two things:

1. Validates checksums that have been provided if a user does that
2. Generates checksums for any and all files that lack the required checksums at this point

There are two types of required checksums at this point in accessioning:

1. md5
2. sha1

If these checksums are not generated by the end of `checksum-compute` then accessioning will eventually fail. Other steps later in the accessionWF rely on having both types of checksums present in the Cocina. The Moab will also require these checksums, plus add sha256 as well.

**User Impact**
When md5 and/or sha1 checksums are missing and the workflow continues on, the resulting error messages can be quite misleading, causing delays in resolving the issue.

For example, a recent occurrence of this error happened during the `shelve` step and made it look like the problem was in preservation. The `shelve` step looked for the required checksums, could not find them, and reported a message implying that the issue had to do with the "content diff" where files being accessioned are compared with files already preserved. But the problem had nothing to do with preservation because the issue occurred on the very first version of the object, which by definition has not yet been preserved.

**To Reproduce**
It's not clear how to reproduce this. Very occasionally, files go through assemblyWF without getting md5 and/or sha1 checksums. Maybe it's a problem more likely to occur when the system is very busy.

**Expected behavior**
`checksum-compute` should compute the required checksums. If that fails, then the step should should report an error. This will allow people to retry from the Argo UI and avoid someone having to go through a bunch of workflow server rake task rigamarole to resolve things.

**Additional context**
We'll have to be careful with validation timing. Some files (such as auto-generated JP2s) will start `checksum-compute` without checksums, so those shouldn't cause an error. The error should only be triggered if checksums are still missing when the step is about to be marked complete.

peetucket:

HOLD - while Andrew investigates if this is actually needed or not

Fixes #1056 - after checksums are computed by the checksum-compute robot, verify that we indeed now have a non-blank MD5 and SHA1 checksum for each file.  Not sure why this happens, but the ticket states that occasionally files make it through this step without having all checksums computed.  This would throw the step in error if this happens again, which can help with debugging and to prevent confusing error messages further down the accessioning workflows.

new unit tests

### DIFF
--- a/lib/robots/dor_repo/assembly/checksum_compute.rb
+++ b/lib/robots/dor_repo/assembly/checksum_compute.rb
@@ -13,6 +13,7 @@ module Robots
 
           cocina_model = assembly_item.cocina_model
           file_sets = compute_checksums(assembly_item, cocina_model)
+          verify_checksums_exist(file_sets)
 
           # Save the modified metadata
           updated = cocina_model.new(structural: cocina_model.structural.new(contains: file_sets))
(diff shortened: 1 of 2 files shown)


jmartin-sul (approved):

peetucket:
We think the actual problem is https://github.com/sul-dlss/pre-assembly/issues/1171 .... closing this in favor of fixing there.

PR #1099, opened 2023-09-15
Rewrite jp2 create robot to handle files not being present.

### LINKED ISSUE #1090: Targeted file updates in Jp2 create
In targeted file updates, the tiff file will not exist in the workspace if it has not changed. The Jp2Create robot expects it to exist so that a JP2 file can be generated at the cocina model updated.

Possible approaches:
* Ignore, tiff files will not be accessioned with Pre-assembly
* Require tiff files to be present so that Jp2Create can be run (i.e., they cannot be used with targeted file updates)
* In Pre-assembly, if a tiff file has not changed then add the cocina file for the JP2 into the structural. When the tiff file is not present, but there is a cocina file for the Jp2 in the structural, the Jp2Create robot does not attempt to generate a new Jp2.

Also needs to be considered how each of these approaches will affect shelving, which probably assumes that the Jp2 file is present.

Thoughts @andrewjbtw ?

justinlittman:
closes #1090

Deciding how, when to create JP2s is very complex

Unit, integration


edsu:
I think this needs a rebase w/ main to bring in latest Cocina updates?

lwrubel (approved):
Approved, but with so much complicated logic, you might want another pair of eyes on it before merging.

lib/robots/dor_repo/assembly/jp2_create.rb
```diff
@@ -24,82 +24,124 @@ def perform_work
...
 
-            if tuples_to_add.present?
-              # Remove existing jp2 nodes
-              files.delete_if { |file| file.fetch(:filename).ends_with?('.jp2') }
-            end
+              # If the file exists and there is a jp2 cocina file, skip it
+              next if File.exist?(filepath) && File.exist?(jp2_filepath) && cocina_jp2_file.present?
+
+              # If the file does not exist and there is a jp2 cocina file, skip it
+              next if !File.exist?(filepath) && cocina_jp2_file.present?
```
ndushay, line 57: would it make sense to check for the cocina_jp2 file before some of these steps?

lib/robots/dor_repo/assembly/jp2_create.rb
```diff
@@ -24,82 +24,124 @@ def perform_work
...
 
-              next unless object_file.jp2able?
+              assembly_image = assembly_image_for(filepath)
+              jp2_filename = jp2_filename_for(filename, assembly_image)
+              jp2_filepath = filepath_for(jp2_filename, assembly_item)
 
-              img = ::Assembly::Image.new(object_file.path)
-              tuples_to_add << [file, img]
-            end
+              cocina_jp2_file = find_cocina_jp2_file(cocina_files)
```
ndushay, line 51: how is the loop working if the loop is on cocina_files, but we check against all cocina_files to find a single cocina_jp2_file?

lib/robots/dor_repo/assembly/jp2_create.rb
```diff
@@ -24,82 +24,124 @@ def perform_work
...
-              # Remove existing jp2 nodes
-              files.delete_if { |file| file.fetch(:filename).ends_with?('.jp2') }
-            end
+              # If the file exists and there is a jp2 cocina file, skip it
+              next if File.exist?(filepath) && File.exist?(jp2_filepath) && cocina_jp2_file.present?
+
+              # If the file does not exist and there is a jp2 cocina file, skip it
+              next if !File.exist?(filepath) && cocina_jp2_file.present?
+
+              # If the files does not exist and there is no jp2 cocina file, get it from preservation
```
ndushay, line 59: files does not exist <-- typo

ndushay (changes requested):
i have some points of confusion and a suggestion.

lib/robots/dor_repo/assembly/jp2_create.rb
justinlittman: This is all of the cocina files in a file set, not all cocina files.

lib/robots/dor_repo/assembly/jp2_create.rb
justinlittman: It's another option, that I'm not sure that any formulation of this logic is going to be clearer than any other. If you had a formulation, I'd be happy to consider.

lib/robots/dor_repo/assembly/jp2_create.rb
justinlittman: Yes, yes it is. Fixed.

lib/robots/dor_repo/assembly/jp2_create.rb
ndushay: I mean the cocina_files object on line 31 and line 34, which is stepping through them, is also looking at the cocina_files same object on line 51 ... inside the loop.

lib/robots/dor_repo/assembly/jp2_create.rb
ndushay: backing away slowly.  There might be a way but I'm not going to spend the time on it.  If someone wants, they can refactor it later.  Working code wins.

ndushay (approved):

Issue #1074, opened 2023-07-14
Analyze whether checksum robot steps can be made more efficient (i.e. take less time to run)


andrewjbtw:
**Background**

There are two steps in the SDR accessioning workflows that can take an extraordinary amount of time when processing large (multi-TB) content. By extraordinary, I mean 1-5 days for just the one robot step. Both steps are steps that check and/or calculate checksums:

`checksum-compute` in the assemblyWF
`sdr-ingest-transfer` in the accessionWF

**Analysis requested**

This ticket is for analysis and investigation into whether these steps are running as efficiently as they can given our hardware and network bandwidth constraints, or whether they can be sped up.

I'm very curious whether these steps are reading each file in a druid more than once. For example, the `checksum-compute` step may carry out multiple tasks:

- generate MD5 checksum for every file
- generate SHA1 checksum for every file
- verify user-supplied MD5 check for every file

If the robot step is reading each file more than once, would it be possible to modify the step to carry out only one read? There's no way to generate checksums with less than one full read of each file, but it may not be necessary to read more than once in order to carry out the step.

**Some evidence that the `checksum-compute` step may be reading files multiple times**

I've been monitoring the progress of `checksum-compute` on a druid with 5 TB of data. The tool I've been using is `lsof`, which lists open files. By grepping for the druid, I get a list of files that are being processed that are associated with that druid.

```bash
$ lsof -o | grep sz929gx7593
ruby      1900060                        lyberadmin    9r      REG               0,57 0x99d4fb4000 161997549 /dor/assembly/sz/929/gx/7593/sz929gx7593/content/sz929gx7593_5_pm.mkv (sf6-webapp:/sul_lyberservices_prod_01)
ruby      1900060 1901089 Timeout        lyberadmin    9r      REG               0,57 0x99d4f90000 161997549 /dor/assembly/sz/929/gx/7593/sz929gx7593/content/sz929gx7593_5_pm.mkv (sf6-webapp:/sul_lyberservices_prod_01)
ruby      1900060 1901090 sidekiq.h      lyberadmin    9r      REG               0,57 0x99d4f98000 161997549 /dor/assembly/sz/929/gx/7593/sz929gx7593/content/sz929gx7593_5_pm.mkv (sf6-webapp:/sul_lyberservices_prod_01)
[...]

PR #1100, opened 2023-09-17
Change compute checksums to digest in single pass.

justinlittman:
refs #1074

Need for speed.

Unit

### DIFF
--- a/lib/robots/dor_repo/assembly/checksum_compute.rb
+++ b/lib/robots/dor_repo/assembly/checksum_compute.rb
@@ -45,10 +45,8 @@ module Robots
           # File is not changing, so use existing checksums
           return if md5_node && sha1_node && !File.exist?(filepath)
 
-          object_file = ::Assembly::ObjectFile.new(filepath)
-
           # compute checksums
-          checksums = { md5: object_file.md5, sha1: object_file.sha1 }
+          checksums = generate_checksums(filepath)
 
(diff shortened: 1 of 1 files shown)


lwrubel (approved):

mjgiarlo:
@justinlittman do you think we should consider extracting this to the assembly-objectfile gem too? Maybe it's worthy of a new method in the public interface of the gem. (Definitely backloggable if so.)

ndushay (commented):
wow - this was a huge oversight.

ndushay:
> @justinlittman do you think we should consider extracting this to the assembly-objectfile gem too? Maybe it's worthy of a new method in the public interface of the gem. (Definitely backloggable if so.)

the main advantage would be computing the two flavors of checksum at the same time if both were desired:  https://github.com/sul-dlss/assembly-objectfile/blob/main/lib/assembly/object_file.rb#L89-L98

It might also be worth figuring out if checksums are actually being computed with the gem.  https://github.com/sul-dlss/assembly-objectfile/network/dependents

mjgiarlo:
@ndushay 💬 
> It might also be worth figuring out if checksums are actually being computed with the gem. [sul-dlss/assembly-objectfile (dependents)](https://github.com/sul-dlss/assembly-objectfile/network/dependents)

Good call. it could be the spot in common-accessioning that was already optimized for speed using inline code is the only spot where we care about such an optimization, so the extraction could be a chunk of work without much payoff.

PR #1111, opened 2023-10-18
Change reset to a noop so DSA can set 'completed' when doing the actual work

justinlittman:
refs https://github.com/sul-dlss/dor-services-app/issues/4634

To prevent the workflow from proceeding until reset is actually completed by DSA job.

Unit, stage

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem 'activesupport', '~> 7.0'
 gem 'assembly-image', '~> 2.0' # ruby-vips is used by 2.0.0 for improved image processing
 gem 'assembly-objectfile', '~> 2.1'
 gem 'config', '~> 2.2'
-gem 'dor-services-client', '~> 13.0'
+gem 'dor-services-client', '~> 13.1'
 gem 'dor-workflow-client', '~> 5.0'
 gem 'dry-struct', '~> 1.0'
 gem 'dry-types', '~> 1.1'
--- a/lib/robots/dor_repo/accession/reset_workspace.rb
+++ b/lib/robots/dor_repo/accession/reset_workspace.rb
@@ -11,7 +11,11 @@ module Robots
         end
 
         def perform_work
-          object_client.workspace.reset
+          # Reset workspace is performed async by dor-services-app.
+          object_client.workspace.reset(workflow: 'accessionWF', lane_id:)
+
+          # dor-services-app will update the workflow step, do don't do it here.
+          LyberCore::ReturnState.new(status: :noop, note: 'Initiated reset API call.')
         end
       end
     end
--- a/spec/robots/dor_repo/accession/end_accession_spec.rb
+++ b/spec/robots/dor_repo/accession/end_accession_spec.rb
@@ -25,9 +25,11 @@ RSpec.describe Robots::DorRepo::Accession::EndAccession do
   describe '#perform' do
     subject(:perform) { test_perform(robot, druid) }
 
+    let(:return_status) { perform.status }
+
     context 'when there is no special dissemniation workflow' do
       it 'cleans up' do
-        perform
+        expect(return_status).to eq 'noop'
         expect(workspace_client).to have_received(:cleanup).with(workflow: 'accessionWF', lane_id: 'default')
       end
     end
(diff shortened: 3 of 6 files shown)


spec/robots/dor_repo/accession/reset_workspace_spec.rb
```diff
@@ -7,20 +7,20 @@
...
+    allow(LyberCore::WorkflowClientFactory).to receive(:build).and_return(workflow_client)
   end
 
   describe '#perform' do
     subject(:perform) { test_perform(robot, druid) }
 
-    before do
-      perform
-    end
-
```
ndushay, line 21: without this ... how is this testing what happens when the reset step is performed?

spec/robots/dor_repo/accession/end_accession_spec.rb
```diff
@@ -27,7 +27,7 @@
 
     context 'when there is no special dissemniation workflow' do
       it 'cleans up' do
-        perform
```
ndushay, line 30: again - doesn't it have to actually perform first?

Gemfile
```diff
@@ -7,7 +7,8 @@ gem 'activesupport', '~> 7.0'
 gem 'assembly-image', '~> 2.0' # ruby-vips is used by 2.0.0 for improved image processing
 gem 'assembly-objectfile', '~> 2.1'
 gem 'config', '~> 2.2'
-gem 'dor-services-client', '~> 13.0'
+# gem 'dor-services-client', '~> 13.0'
+gem 'dor-services-client', github: 'sul-dlss/dor-services-client', branch: 'reset'
```
ndushay, line 11: this clearly isn't meant to be merged yet.

ndushay (changes requested):

lib/robots/dor_repo/accession/reset_workspace.rb
```diff
@@ -11,7 +11,8 @@ def initialize
         end
 
         def perform_work
-          object_client.workspace.reset
+          object_client.workspace.reset(workflow: 'accessionWF', lane_id:)
```
ndushay, line 14: so this is calling an async job to be run??  Maybe a comment could help

spec/robots/dor_repo/accession/end_accession_spec.rb
```diff
@@ -27,7 +27,7 @@
...
+        expect(perform.status).to eq 'noop'
```
ndushay, line 30: or does "perform.status" first do the perform and then return the status?   because this is a sort of method name chain?

Gemfile
ndushay: if this needs to be here, even temporarily, maybe a comment explaining that we can't cut gem release until xxx happens?

spec/robots/dor_repo/accession/end_accession_spec.rb
jmartin-sul: yeah, i think that's what's happening.  i believe the "status" here is the lybercore "noop" code indicating that the robot did none of the actual desired work, but delegated it to another service, which will then mark the workflow step complete once the work's actually been performed.  otherwise, by default, once `perform`/`perform_work` returns, the workflow step would be marked complete (via some lybercore magic?), when it's more likely that DSA hasn't actually done the work yet.

confusing because it looks similar to `ActiveJob`'s `perform` and `perform_later` methods, the latter of which wouldn't give you any status info other than whether the job had been queued or not.

i think the goal here is to do something similar to what we do in the pres bots `validate-moab` handoff to a pres cat worker, or the [common-accessioning handoff to DSA](https://github.com/sul-dlss/common-accessioning/pull/1081) to do the actual work for the `update-marc` step.  both of those use this `noop` pattern.

spec/robots/dor_repo/accession/reset_workspace_spec.rb
jmartin-sul: looks like it now gets called inline in the test.  i guess the other option would've been to replace the `before` with something like `let(:status) { perform }` and then `expect` on `status` instead of `perform.status`.  i might like what's here slightly better than that, but i think those options are similarly clear.

ndushay:
linting error, btw.

Gemfile
justinlittman: Sorry @ndushay. This PR should have been a draft.

lib/robots/dor_repo/accession/reset_workspace.rb
justinlittman: Comment added.

spec/robots/dor_repo/accession/end_accession_spec.rb
justinlittman: Refactored to make this a tad clearer.

spec/robots/dor_repo/accession/reset_workspace_spec.rb
justinlittman: Refactored to make this a tad clearer.

jmartin-sul (approved):

ndushay (approved):

PR #1132, opened 2024-01-05
Changes to release members robot

justinlittman:
…ollections.

* The members endpoint no longer limits to published items.
* Collections can't be members.


Gemfile
```diff
@@ -7,7 +7,8 @@ gem 'activesupport', '~> 7.0'
 gem 'assembly-image', '~> 2.0' # ruby-vips is used by 2.0.0 for improved image processing
 gem 'assembly-objectfile', '~> 2.1'
 gem 'config', '~> 2.2'
-gem 'dor-services-client', '~> 13.4'
```
justinlittman, line 10: This will be fixed before merge.

lib/robots/dor_repo/release/release_members.rb
```diff
@@ -17,16 +17,19 @@ def perform_work
...
-          publish_collection
+          members.each { |member| create_release_workflow(member) }
         end
 
         private
 
-        def publish_collection
-          member_service = Dor::Release::MemberService.new(druid:)
-          add_workflow_to_members(member_service.items) if add_wf_to_members?
-          add_workflow_to_members(member_service.sub_collections)
```
justinlittman, line 29: This all becomes easier because removing code for releasing sub_collections. Collections can't have collections as members.

lib/robots/dor_repo/release/release_members.rb
```diff
@@ -17,16 +17,19 @@ def perform_work
...
+        def members
+          object_client.members.select { |member| published?(member) }
```
justinlittman, line 28: The published check was previously performed by DSA. However, it is not performed here as it may take some time (since getting from workflow service instead of solr).

lib/robots/dor_repo/release/release_members.rb
```diff
@@ -39,26 +42,8 @@ def add_wf_to_members?
...
-            create_release_workflow(sub_collection['druid'])
-          end
-        end
-
-        def create_release_workflow(druid)
-          logger.debug "...adding workflow releaseWF for #{druid}"
-          object_client = Dor::Services::Client.object(druid)
-
-          # initiate workflow by making workflow service call
-          current_version = object_client.version.current
```
justinlittman, line 60: The version is now provided by the members endpoint, so no need to fetch here.

mjgiarlo (approved):

PR #1161, opened 2024-04-01
not using sul_orcid_client in this code?

peetucket:

I believe we are not using the `sul_orcid_client` gem in this code (we may have before but now it is in DSA).  I don't see it being used anywhere in the code and it doesn't appear to be configured here either.

Here is an example of where it is configured in DSA: https://github.com/sul-dlss/dor-services-app/blob/main/app/jobs/update_orcid_work_job.rb#L39-L47 and no equivalent references in common-accessioning: https://github.com/search?q=repo%3Asul-dlss%2Fcommon-accessioning%20SulOrcidClient&type=code

CI

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,6 @@ gem 'pry'
 gem 'rake'
 gem 'sidekiq', '~> 7.0'
 gem 'slop'
-gem 'sul_orcid_client'
 gem 'zeitwerk', '~> 2.1'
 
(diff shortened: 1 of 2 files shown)


edsu (approved):

peetucket:
@justinlittman ^^ in case this rings a bell and you wanted to verify that the sul_orcid_client is not needed in common accessioning

PR #1170, opened 2024-04-05
Changes accessioning initiate robot to close version.

### LINKED ISSUE #1162: AccessioningInitiate robot should close version
Instead of initiating an accessioning workflow.

justinlittman:
closes #1162

Versioning normalization.

Unit, integration

### DIFF
--- a/lib/robots/dor_repo/assembly/accessioning_initiate.rb
+++ b/lib/robots/dor_repo/assembly/accessioning_initiate.rb
@@ -14,7 +14,7 @@ module Robots
         def perform_work
           logger.info("Initiate accessioning for #{druid}")
           initialize_workspace if assembly_item.item?
-          start_accession_workflow
+          close_version
           true
         end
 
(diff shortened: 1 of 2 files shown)


jcoyne (commented):
Shouldn't a version already be closed before the workflow is started?  We have lots of documentation saying this is so.  Do we need to ADR this before making this change? https://consul.stanford.edu/display/DIGMETADATA/Opening+and+closing+object+versions+in+Argo

lib/robots/dor_repo/assembly/accessioning_initiate.rb
```diff
@@ -24,9 +24,8 @@ def initialize_workspace
           object_client.workspace.create(source: assembly_item.path_finder.path_to_object)
         end
 
-        def start_accession_workflow
-          current_version = object_client.version.current
-          workflow_service.create_workflow_by_name(druid, 'accessionWF', version: current_version, lane_id:)
+        def close_version
+          object_client.version.close
```
jcoyne, line 28: Should we check the return value here? What if this can't be closed?

jcoyne (changes requested):

lib/robots/dor_repo/assembly/accessioning_initiate.rb
justinlittman: If it can't be closed, then an exception will be raised and the workflow step will fail.

I'm not seeing the loop. This is for the assemblyWF to close a version, which will initiate accessionWF.

jcoyne: Got it. I was thinking this was accessioningWF.

jcoyne (approved):

PR #1180, opened 2024-04-17
add new robots for captioning and ocr

### LINKED ISSUE #1168: Write end-ocr and end-caption robots
Closes the version

### LINKED ISSUE #1169: Write start-ocr and start-caption robots
Opens the new version. Needs to verify that the version is openable to avoid race conditions since both of these workflows can be started (or they can be started when the object is being edited elsewhere).  If already open, will remain queued.

### LINKED ISSUE #1179: Create starter robots for new workflows
Write shell robots for new workflows that will be filled in later.

peetucket:

Fixes #1168 and fixes #1169 and fixes #1179 - create starter robots for new captionWF and ocrWF

~~Waiting on sul-dlss/workflow-server-rails#757~~

Functionality to come later.

New stub specs

### DIFF
--- a/spec/robots/dor_repo/caption/end_caption_spec.rb
+++ b/spec/robots/dor_repo/caption/end_caption_spec.rb
@@ -6,7 +6,19 @@ describe Robots::DorRepo::Caption::EndCaption do
   let(:druid) { 'druid:bb222cc3333' }
   let(:robot) { described_class.new }
 
-  it 'runs the end caption robot' do
+  let(:object) { build(:dro, id: druid) }
+  let(:workspace_client) { instance_double(Dor::Services::Client::Workspace, create: nil) }
+  let(:version_client) { instance_double(Dor::Services::Client::ObjectVersion, close: true) }
+  let(:object_client) do
+    instance_double(Dor::Services::Client::Object, version: version_client, workspace: workspace_client, find: object)
+  end
+
+  before do
+    allow(Dor::Services::Client).to receive(:object).and_return(object_client)
+  end
+
+  it 'closes the object' do
     expect(test_perform(robot, druid)).to be true
+    expect(version_client).to have_received(:close)
   end
 end
--- a/spec/robots/dor_repo/caption/end_caption_spec.rb
+++ b/spec/robots/dor_repo/caption/end_caption_spec.rb
@@ -7,7 +7,7 @@ describe Robots::DorRepo::Caption::EndCaption do
   let(:robot) { described_class.new }
 
   let(:object) { build(:dro, id: druid) }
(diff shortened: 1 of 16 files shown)


spec/robots/dor_repo/caption/end_caption_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Robots::DorRepo::Caption::EndCaption do
+  let(:druid) { 'druid:bb222cc3333' }
+  let(:robot) { described_class.new }
+
+  let(:object) { build(:dro, id: druid) }
+  let(:workspace_client) { instance_double(Dor::Services::Client::Workspace, create: nil) }
```
edsu, line 10: I noticed that `create: nil` isn't needed here, I guess because `create()` doesn't actually get called? If we don't expect it to be called I suggest we remove this since it implies that it will.

spec/robots/dor_repo/caption/start_caption_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe Robots::DorRepo::Caption::StartCaption do
+  let(:druid) { 'druid:bb222cc3333' }
+  let(:robot) { described_class.new }
+
+  let(:object) { build(:dro, id: druid) }
+  let(:workspace_client) { instance_double(Dor::Services::Client::Workspace, create: nil) }
```
edsu, line 10: I have the same question here: should `create()` be getting called, because it appears that the tests still pass when removing `create: nil`.

edsu (approved):
I left some questions, which I would like to know the answer to for my own understanding, but they aren't critical since this is all stubbed out.

spec/robots/dor_repo/caption/start_caption_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    it 'opens the object' do
+      expect(test_perform(robot, druid)).to be true
+      expect(version_client).to have_received(:open)
+    end
+  end
+
+  context 'when the object is already opened' do
+    let(:version_open) { true }
+
+    it 'does nothing' do
```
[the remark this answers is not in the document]
edsu: But don't we want to prevent the ocr and caption workflows from running if the item is already open?

spec/robots/dor_repo/caption/start_caption_spec.rb
peetucket: yes, good call... i removed the mocked method in the new tests where it is not needed

PR #1196, opened 2024-05-06
start ocr logic for end-accession

### LINKED ISSUE #1163: Add logic to determine if OCR/caption is needed
Possibly a new class in the lib/dor .

This logic will be used by the end-accession robot to decide if we need to initialize the new `textGenerationWF`.

See https://docs.google.com/document/d/1JLJwio7xVDDh75KY3dZJIFPDep-92hoQJ5p9EgFKabY/edit#heading=h.n930qglopzgc

and https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.b6tux0qgfzvx

Will also need to inspect the content type and a new workflow variable (requiresOCR) to decide if the workflow is needed.

The new workflow will handle both OCR and transcription since some objects may have both needed.

### LINKED ISSUE #1164: Initialize ocr or caption workflows from end-accession
Using logic from #1163, initialize ocr or caption workflows if needed from end-accession

peetucket:

Fixes #1163 and fixes #1164

The end-accession robot will start the ocrWF as needed.

~~- check for metadata in cocina to determine if we need to start the ocrWF~~ deferred to later ticket #1203

~~Currently blocked by sul-dlss/cocina-models#705 until will resolve the cocina models decisions~~

Notes: 

- the logic for starting OCR workflow is not yet complete, we need to deal with the checks for cocina metadata still.
- captionWF is ignored for now, will add later when we actually work on it

### DIFF
--- a/lib/dor/text_extraction/ocr.rb
+++ b/lib/dor/text_extraction/ocr.rb
@@ -12,11 +12,13 @@ module Dor
       end
 
       def required?
-        return false unless cocina_object.dro? # only items can be OCR'd
+        # only items can be OCR'd
+        return false unless cocina_object.dro?
 
+        # only specific content types can be OCR'd
         return false unless allowed_object_types.include? cocina_object.type
 
-        # user has indicated that OCR should be run (set as workflow_context)
+        # checks if user has indicated that OCR should be run (set as workflow_context)
         return false unless workflow_context['runOCR']
 
         # TODO: check for required input files for OCR? (skip if none found) e.g. by mimetype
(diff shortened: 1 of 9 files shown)


lib/dor/text_extraction/ocr.rb
```diff
@@ -0,0 +1,72 @@
...
+        return false unless allowed_object_types.include? cocina_object.type
+
+        # checks if user has indicated that OCR should be run (set as workflow_context)
+        return false unless workflow_context['runOCR']
+
+        # check to be sure there are actually files to be OCRed
+        return false unless filenames_to_ocr.any?
+
+        # TODO: check for any files that have "manuallyCorrected" in cocina structural (then skip)
+
```
peetucket, line 28: raise an error for cases where the user asked for OCR but it is not possible instead of just silently continuing

dnoneill (approved):

PR #1200, opened 2024-05-07
functions to build abby ticket for processing

### LINKED ISSUE #1189: Create XML ticket file specifying input file location and desired output file locations
part of https://github.com/sul-dlss/common-accessioning/issues/1166 (ocrWF:ocr-create)

ABBYY needs an XML file to specify OCR params and the files to operate on.  Create this file and place in the ABBYY watch folder

See https://help.abbyy.com/en-us/finereaderserver/14/help/xmlticket_description/

and https://docs.google.com/document/d/1kjcqMUi5HehLOS-tD78CpTUcPN2CAHe-nqSp7WfWZNQ/edit#heading=h.fwuwrf2f5d34

and example jobs: https://drive.google.com/drive/u/1/folders/1yloq1OlUk6quq4HAPpn-mAJwLf4dCpKN

The ABBYY XMLTicket setup for testing had two folders:

EXCEPTIONS/
INPUT/XMLTICKET
OUTPUT/
RESULT/

To start a job, we put source files + the xmlticket in INPUT/XMLTICKET

If all went well, ABBYY created files in OUTPUT resultxml (like a log) went into RESULT; source files (including xmlticket) were deleted.

If there was a problem, a file was created in EXCEPTIONS.

ABBYY won’t start processing until all the files listed in the xmlticket have been dropped into the folder, one of the advantages over their standard “watch” folder which might start consuming files before you’re ready. And there’s an option to set the output folder in the xmlticket.

Organize input and output by druid if possible.

dnoneill:
closes #1189 

- Via created tests in rspec

### DIFF
--- a/lib/dor/text_extraction/abbyy.rb
+++ b/lib/dor/text_extraction/abbyy.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+module Dor
+  module TextExtraction
+    # Build Abbyy tickets for processing
+    class Abbyy
+      def initialize(filepaths:, druid:)
+        @filepaths = filepaths
+        @druid = druid
+      end
+
+      def abbyy_ticket_filepath
+        File.join(Settings.sdr.abbyy_ticket_path, "#{@druid}.xml")
+      end
+
+      def image_abbyy
+        [pdfa_output, alto_output, text_output].join("\n")
+      end
+
+      def pdf_files?
+        File.extname(@filepaths.first).strip.downcase[1..] == 'pdf'
+      end
+
+      def alto_output
+        "<ExportFormat OutputFileFormat='ALTO' OutputFlowType='SharedFolder' FormatVersion='3_1' CoordinatesParticularity='Words' WriteWordConfidence='true'>
+          <OutputLocation>#{Settings.sdr.abbyy_output_path.encode(xml: :text)}</OutputLocation>
+          <FileExistsAction>Overwrite</FileExistsAction>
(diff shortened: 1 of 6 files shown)


edsu (commented):
Just to be on the safe side should we xml encode the element and attribute text? Otherwise it's possible we could send ABBYY invalid XML?

```ruby
element_text.encode(xml: :text)
attribute_text.encode(xml: :attr)
```

dnoneill:
Worked great. Thanks for the suggestion @edsu

edsu (approved):

lib/dor/text_extraction/abbyy.rb
```diff
@@ -0,0 +1,79 @@
...
+      def abbyy_ticket_filepath
+        File.join(Settings.sdr.abbyy_ticket_path, "#{druid}.xml")
+      end
+
+      def image_abbyy
+        [pdfa_output, alto_output, text_output].join("\n")
+      end
+
+      def pdf_files?
+        File.extname(filepaths.first).strip.downcase[1..] == 'pdf'
```
edsu, line 29: I'm not totally up on the logic here, but do we want to look at all the file paths to see if there is a PDF?

edsu (approved):

lib/dor/text_extraction/abbyy.rb
dnoneill: I am assuming the filtering of files will happen before we build the ticket and they should all be the same type.

edsu: That is probably worth a comment, since it is confusing.

Issue #1191, opened 2024-05-01
Watch ABBYY output folder to determine when OCR is complete


peetucket:
part of https://github.com/sul-dlss/common-accessioning/issues/1166 (ocrWF:ocr-create)

We need a process that is continually running that monitors the ABBYY folder to determine when an OCR is complete for an object so that the workflow step can be updated appropriately.

thatbudakguy:
a quick survey of available tech for this was not super encouraging.

my goal was to see if there are any system-level tools for checking this sort of thing _without_ polling, so that we wouldn't have the overhead of lots of processes continually running checking to see if something in the folder changed.

it seems that the way this is normally done on linux systems (presuming that's what common-accessioning is) is through [`inotify`](https://man7.org/linux/man-pages/man7/inotify.7.html), which most of the GUI linux filemanagers (GNOME, etc.) use under the hood.

there is a lot of discussion out there about how `inotify` doesn't play nice with network fileystems, including samba. this [transcript of a linux foundation meeting](https://lwn.net/Articles/896055/) is basically lamenting how this is still the case in 2022, where OSX and windows systems can receive change notifications about network filesystems but linux is still behind. see also this [thread on the docker forums](https://forums.docker.com/t/file-system-watch-does-not-work-with-mounted-volumes/12038) about how monitoring filesystems from docker containers suffers the same problem for the same reason.

systemd provides [the path unit](https://www.freedesktop.org/software/systemd/man/latest/systemd.path.html), which you can use to set up all kinds of automatic notifications about filesystems as a background process. unfortunately, it uses `inotify` too:
> Internally, path units use the [inotify(7)](https://man7.org/linux/man-pages/man7/inotify.7.html) API to monitor file systems. Due to that, it suffers by the same limitations as inotify, and for example cannot be used to monitor files or directories changed by other machines on remote NFS file systems.

in ruby-land, the [`listen` gem](https://github.com/guard/listen) looks like a reasonably well-supported way to do this and provides a nice API, but it too warns:
> Some filesystems won't work without polling (VM/Vagrant Shared folders, NFS, Samba, sshfs, etc.).

that said, changing to use polling is just a single option in its API (`force_polling: true`), so at least it's not hard to turn on if we have to.

thatbudakguy:
the [samba docs](https://wiki.samba.org/index.php/Editing_files_locally_on_server:_interoperability) make it sound like change notifications are possible, but I don't totally understand it:

[...]

PR #1205, opened 2024-05-10
parse abbyy xml output success, files, errors

### LINKED ISSUE #1190: Parse ABBYY output xml file
part of https://github.com/sul-dlss/common-accessioning/issues/1166 (ocrWF:ocr-create)

ABBYY produces a result.xml file in the output folder when OCRing is complete.  We should parse this file to update the status of the workflow step appropriately

Notes:
- The output files will go to where they're defined in the element--which wouldbe used for the alto XML file. 
- The RESULTXML has been the location used for the Abbyy "result" file that shows everything that happened to the file--this would include information about any errors that might have happened. The result xml will also indicate if there is a failure or not in the job.
- The top element in the result.xml has a IsFailed="true" when there is a problem (or false when there isn't).

dnoneill:
Closes #1190 

This PR makes a couple of assumptions.
1. The watcher in #1191 is going to have a filepath that includes the results.xml full filepath. i.e. `/abby/RESULTXML/druid.xml.result.xml` or `/abby/EXCEPTIONS/druid.xml.result.xml`. The reason I am thinking the watcher will want to be the determiner is if the same druid has been run more than once the .xml.result.xml file gets a number prepended, i.e. druid.0001.xml.result.xml which will be easier to spot with a watcher than code.

2. We want the errors Abbyy provides for the interface
3. We want the output files if Abbyy was successful.

### Major Changes
1. Moved what used to be the Abbyy class into a subclass called Ticket. Created a new class called Results for this work.

Tested via rspec. Test files were pulled from the qa server after Abbyy tests.


lib/dor/text_extraction/abbyy.rb
```diff
@@ -2,77 +2,105 @@
...
+          @filepaths = filepaths
+          @druid = druid
+        end
 
-      def write_xml_ticket
-        File.write(abbyy_ticket_filepath, xml_ticket)
-      end
+        def write_xml_ticket
+          File.write(abbyy_ticket_filepath, xml_ticket)
+        end
```
thatbudakguy, line 17: could we simplify these method names now that this is called `Ticket`? like this could be `write_xml`, `abbyy_ticket_filepath` could just be `file_path`, etc. (which is a sign that this is a good change, imo!)

dnoneill: That makes sense. Thanks. Done.

thatbudakguy (approved):
thanks!

peetucket (approved):

PR #1209, opened 2024-05-14
update ocr to filter files

### LINKED ISSUE #1201: Determine which files to add to OCR XML ticket
In order to create the XML ticket to send to ABBYY, we need to know which files will be OCRed.

This may be a simple list of files that match specific mimetypes, but will require some logic, as there will be cases where the same content is in two different formats (e.g. Jp2 and Tiff).  These should be bundled into the same resource, so maybe just pick one in a given resource in some preferential order?

related discussion in https://github.com/sul-dlss/common-accessioning/issues/1187#issuecomment-2101041870

dnoneill:
- filter files for OCR based on type and filetypes.

fixes #1201 

RSPEC. ocr_file function tested on the server in console with `druid:qv402bt5465` and a type document druid.

### DIFF
--- a/lib/dor/text_extraction/ocr.rb
+++ b/lib/dor/text_extraction/ocr.rb
@@ -51,6 +51,8 @@ module Dor
         end
       end
 
+      # filter down fileset files to those in preservation and are allowedmimetypes
+      # if there are more than one allowed mimetype, grab the preferred type
       def ocr_file(fileset)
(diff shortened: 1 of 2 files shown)


peetucket (approved):
Added code suggestion as a new commit.

peetucket:
At some point (maybe now, maybe later?) we could add some more specs around the `ocr_files` class to make sure it does the expected thing for other variations of content types and/or non-preserved files.

PR #1210, opened 2024-05-14
Add fetch_files for OCR

### LINKED ISSUE #1187: Copy files to be OCRed from preservation
part of #1166 (ocrWF:ocr-create)

Blocked by #1188
Since we will be running ocrWF after accessionWF has been run, all of the files have already been cleared off the workspace.  In order to run OCR, we will need to get the files from preservation and put them in the ABBYY watch folder.

So we need to fetch files from preservation and move them to ABBYY

edsu:

Add a workflow step for wfOCR to fetch files in need of OCR from preservation using the Preservation::Client.

Closes #1187

Unit tests. It's not actually part of the workflow yet...

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,12 +37,13 @@ end
 group :development do
   gem 'capistrano', '~> 3.0'
   gem 'capistrano-bundler', '~> 1.1'
+  gem 'debug'
   gem 'dlss-capistrano', require: false
+  gem "ruby-debug-completion"
 end
 
 group :development, :test do
-  gem 'byebug'
-  gem 'pry-byebug', platform: %i[ruby_20 ruby_21]
   gem 'rubocop'
   gem 'rubocop-rspec'
 end
+
--- a/lib/dor/text_extraction/ocr.rb
+++ b/lib/dor/text_extraction/ocr.rb
@@ -31,14 +31,14 @@ module Dor
         workflow_context['runOCR'] || false
       end
 
-      private
-
       # return a list of filenames that should be OCR'd
       # iterate over all files in cocina_object.structural.contains, looking at mimetypes
       # return a list of filenames that are correct mimetype
       def filenames_to_ocr
         ocr_files.map(&:filename)
       end
+
+      private
 
       # iterate through cocina strutural contains and return all File objects for files that need to be OCRed
       def ocr_files
--- a/lib/robots/dor_repo/ocr/fetch_files.rb
+++ b/lib/robots/dor_repo/ocr/fetch_files.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module Robots
+  module DorRepo
+    module Ocr
+      # Fetch files in need of OCR from Preservation
+      class FetchFiles < LyberCore::Robot
+        def initialize
+          super('ocrWF', 'ocr-fetch-files')
+        end
+
+        def perform_work
+          copy_files
+        end
+
+        private
+
+        def copy_files
+          ocrable_filenames.each do |filename|
+            path = abbyy_path(filename)
+            path.parent.mkpath unless path.parent.directory?
+            path.open('wb') do |file_writer|
+              preservation_client.objects.content(druid:,
+                                                  filepath: filename,
+                                                  on_data: proc { |data, _count| file_writer.write(data) })
+            end
+          end
+        end
+
+        def bare_druid
+          druid.split(':').last
+        end
+
+        def abbyy_path(filename)
+          # NOTE: if files of type "file" were allowed here we would have to
+          # deal with file hierarchy (subdirectories)
+          Pathname.new(File.join(Settings.sdr.abbyy_ticket_path, bare_druid, filename))
+        end
+
+        def ocrable_filenames
+          @ocrable_filenames ||= ocr.required? && ocr.possible? ? ocr.filenames_to_ocr : []
+        end
+
+        def ocr
+          @ocr = Dor::TextExtraction::Ocr.new(cocina_object:, workflow_context: workflow.context)
+        end
+
+        def preservation_client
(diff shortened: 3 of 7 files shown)


Gemfile
```diff
@@ -37,12 +37,13 @@ end
 group :development do
   gem 'capistrano', '~> 3.0'
   gem 'capistrano-bundler', '~> 1.1'
+  gem 'debug'
   gem 'dlss-capistrano', require: false
+  gem "ruby-debug-completion"
 end
 
 group :development, :test do
-  gem 'byebug'
```
edsu, line 44: From Slack it seemed like 'debug' was the future 'byebug' was the past? If you'd rather keep it I can put it back.

lib/dor/text_extraction/ocr.rb
```diff
@@ -31,15 +31,15 @@ def required?
...
-      private
-
       # return a list of filenames that should be OCR'd
       # iterate over all files in cocina_object.structural.contains, looking at mimetypes
       # return a list of filenames that are correct mimetype
       def filenames_to_ocr
         ocr_files.map(&:filename)
       end
 
+      private
```
edsu, line 41: I needed filenames_to_ocr to be public so that the fetch_files step would know what files to get from Preservation.

spec/robots/dor_repo/ocr/fetch_files_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+  end
+
+  let(:pres_client) do
+    instance_double(Preservation::Client, objects: objects_client)
+  end
+
+  let(:objects_client) do
+    instance_double(Preservation::Client::Objects)
+  end
+
```
edsu, line 105: I needed to mock out workflow_client and other things following in order for the `workflow.context` step to work. If there's an easier way around this let me know. Also, we may need similar code in other workflow tests, which might make copy & pasting this a bit tedious?

spec/robots/dor_repo/ocr/fetch_files_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+      it 'gets the files' do
+        perform
+        expect(objects_client).to have_received(:content).twice
+        expect(File.exist?(File.join(Settings.sdr.abbyy_ticket_path, 'image111.tif'))).to be true
+        expect(File.exist?(File.join(Settings.sdr.abbyy_ticket_path, 'image112.tif'))).to be true
+      end
+    end
+  end
+end
```
edsu, line 154: Note I didn't really test differently shaped Cocina since it seemed like that's the job of `spec/lib/dor/text_extraction/ocr_spec.rb` ?

Gemfile
peetucket: i lost track and will go with the flow

lib/robots/dor_repo/ocr/fetch_files.rb
```diff
@@ -0,0 +1,46 @@
...
+                                                  on_data: proc { |data, _count| file_writer.write(data) })
+            end
+          end
+        end
+
+        def abbyy_path(filename)
+          # NOTE: if files of type "file" were allowed here we would have to
+          # deal with file hierarchy (subdirectories)
+          File.join(Settings.sdr.abbyy_ticket_path, filename)
+        end
```
peetucket, line 32: We are going to organize the files by druid, so that the folder structure will be something like:

`/abbyy/INPUT/ab123df4567/files...`

So I think this should be something like:

```
 File.join(Settings.sdr.abbyy_ticket_path, druid, filename)
```

And I think we need to create that druid subfolder if it doesn't exist.  So something like this at the start of `#copy_files`:

```
FileUtils.mkdir(File.join(Settings.sdr.abbyy_ticket_path, druid)) unless unless File.directory?(File.join(Settings.sdr.abbyy_ticket_path, druid))
```

But I guess to avoid all the code duplication, could add this private method instead and then replace all the duplicated code above with the method:

```
def base_abbyy_folder
  File.join(Settings.sdr.abbyy_ticket_path, druid)
end
```

peetucket: Which reminds me, we probably need a new "cleanup abbyy folder" robot at the end to clear out the druid folder.  Even though ABBYY clears out the files themselves after OCRing, i think the empty folder will remain.  I'll ticket that.

peetucket (changes requested):

lib/robots/dor_repo/ocr/fetch_files.rb
edsu: > We are going to organize the files by druid, so that the folder structure will be something like:
> 
> `/abbyy/INPUT/ab123df4567/files...`

It looks like the ticket is getting written to `/abbyy/INPUT/` https://github.com/sul-dlss/common-accessioning/blob/main/lib/dor/text_extraction/abbyy.rb#L21-L23

And the file paths in the ticket reference the files as being alongside the ticket: https://github.com/sul-dlss/common-accessioning/blob/main/lib/dor/text_extraction/abbyy.rb#L64-L66

Should the ticket still get written to that location but reference the file in the druid subdirectory?

peetucket: I think the way this XML writing code is currently written, it is assuming you will pass in an array of filenames where the filepaths include the "druid" subfolder each time.  This is the most flexible but perhaps not ideal.  Maybe we need to put this kind of logic (i.e. the "base folder" for the files) in one place (e.g. in the `Ocr` class) and then use it in both places?

edsu: I've modified how this works over in #1215 so that the caller just knows the filenames (possible hierarchical) and the Abbyy class takes care of naming things with DOS path separator, etc.

edsu: The code should be updated now to write to a druid subdirectory.

peetucket (approved):

PR #1215, opened 2024-05-15
Handle referencing files in druid subdirectory

edsu:

Rather than requiring the user of the Abbyy class to know that file
paths should be in a druid subdirectory using Windows file separator the
caller can just pass in the filenames and the Abbyy class will handle
making sure they are referenced in the XML using the druid subdirectory.

Also update the paths used so that they include the "bare" druid,
or only the identifier only includes the identifier portion, so
instead of '/tmp/druid:bb22ccc3333' it will be '/tmp/bb22ccc3333'.

Even though they are currently out of scope the ticket writing will also
handle when there is file hierarchy in the path: 'x/y.png' instead of
'y.png'.

Unit tests.

### DIFF
--- a/lib/dor/text_extraction/abbyy/ticket.rb
+++ b/lib/dor/text_extraction/abbyy/ticket.rb
@@ -16,10 +16,14 @@ module Dor
           File.write(file_path, xml)
         end
 
+        def file_path
+          File.join(Settings.sdr.abbyy_ticket_path, "#{bare_druid}.xml")
+        end
+
         private
 
-        def file_path
-          File.join(Settings.sdr.abbyy_ticket_path, "#{druid}.xml")
+        def bare_druid
+          @bare_druid ||= druid.split(':').last
         end
 
         def image_output
@@ -31,14 +35,14 @@ module Dor
         end
 
         def output_file_path
-          File.join(Settings.sdr.abbyy_output_path, druid).encode(xml: :text)
+          File.join(Settings.sdr.abbyy_output_path, bare_druid).encode(xml: :text)
         end
 
         def alto_output
           "<ExportFormat OutputFileFormat='ALTO' OutputFlowType='SharedFolder' FormatVersion='3_1' CoordinatesParticularity='Words' WriteWordConfidence='true'>
             <OutputLocation>#{output_file_path}</OutputLocation>
             <FileExistsAction>Overwrite</FileExistsAction>
-            <NamingRule>#{druid}.&lt;Ext&gt;</NamingRule>
+            <NamingRule>#{bare_druid}.&lt;Ext&gt;</NamingRule>
           </ExportFormat>"
         end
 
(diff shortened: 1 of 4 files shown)


lib/dor/text_extraction/abbyy/ticket.rb
```diff
@@ -16,10 +16,14 @@ def write_xml
           File.write(file_path, xml)
         end
 
+        def file_path
```
edsu, line 19: the file path for the ticket seemed reasonable to make public, and simplified the test a bit

lib/dor/text_extraction/abbyy/ticket.rb
```diff
@@ -16,10 +16,14 @@ def write_xml
...
+          File.join(Settings.sdr.abbyy_ticket_path, "#{bare_druid}.xml")
+        end
+
         private
 
-        def file_path
-          File.join(Settings.sdr.abbyy_ticket_path, "#{druid}.xml")
+        def bare_druid
```
edsu, line 25: This is here in case someone happens to pass in a non-bare druid: `druid:...`

lib/dor/text_extraction/abbyy/ticket.rb
```diff
@@ -57,12 +65,17 @@ def pdfa_output
                         PdfUACompatible='true'>
             <OutputLocation>#{output_file_path}</OutputLocation>
             <FileExistsAction>Overwrite</FileExistsAction>
-            <NamingRule>#{druid}.&lt;Ext&gt;</NamingRule>
+            <NamingRule>#{bare_druid}.&lt;Ext&gt;</NamingRule>
           </ExportFormat>"
         end
```
edsu, line 71: The change to `input_filepaths_field` is to let it handle the windows style path for the input file, instead of requiring the user of Abbyy to do it.

lib/dor/text_extraction/abbyy/ticket.rb
```diff
@@ -57,12 +65,17 @@ def pdfa_output
...
 
         def input_filepaths_field
-          filepaths.map { |filename| "<InputFile Name=#{filename.encode(xml: :attr)}/>" }.join("\n")
+          # windows filepath since Abbyy service runs on a Windows machine
+          filepaths.map do |filename|
```
edsu, line 74: Handle potential hierarchy in the files.

lib/dor/text_extraction/abbyy/ticket.rb
```diff
@@ -30,23 +34,27 @@ def pdf_files?
           File.extname(filepaths.first).strip.downcase[1..] == 'pdf'
         end
 
+        def input_file_path
+          File.join(Settings.sdr.abbyy_ticket_path, bare_druid)
+        end
+
         def output_file_path
```
peetucket, line 41: Is this method used externally by someone else?  It doesn't appear to  used in this class so could be removed if not.  If it is used elsewhere, it could then be used below in `#output_file_path` to avoid code duplication.

edsu: I was going to use it in the fetch-files workflow step. but in hindsight it's more trouble than it's worth

dnoneill (approved):

PR #1216, opened 2024-05-15
Create ABBYY file watcher

thatbudakguy:
- Create a workflow updater service for text extraction
- Create a FileWatcher that emits events based on ABBYY jobs

#1231 will create a background service to run the `FileWatcher` using systemd.

### DIFF
--- a/lib/dor/text_extraction/abbyy/file_watcher.rb
+++ b/lib/dor/text_extraction/abbyy/file_watcher.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+module Dor
+  module TextExtraction
+    module Abbyy
+      # Watch jobs ABBYY is processing via the filesystem and report back to SDR
+      class FileWatcher
+        attr_reader :result_xml_path, :exceptions_path
+
+        # These methods control the underlying listener; see:
+        # https://github.com/guard/listen?tab=readme-ov-file#pause--start--stop
+        # NOTE: You need to `sleep` or otherwise avoid letting the process
+        # exit after starting in order for the listener's thread to keep running
+        delegate :start, :pause, :stop, to: :listener
+
+        def initialize(workflow_updater: nil, listener_options: {})
+          @result_xml_path = Settings.sdr.abbyy_result_path
+          @exceptions_path = Settings.sdr.abbyy_exception_path
+
+          # Ensure the ABBYY directories exist
+          raise ArgumentError, "ABBYY result XML path '#{result_xml_path}' is not a directory" unless File.directory?(result_xml_path)
+          raise ArgumentError, "ABBYY exceptions path '#{exceptions_path}' is not a directory" unless File.directory?(exceptions_path)
+
+          # Set up the workflow updater and listener
+          @workflow_updater = workflow_updater || Dor::TextExtraction::WorkflowUpdater.new
+          @listener_options = default_listener_options.merge(listener_options)
+        end
+
+        private
+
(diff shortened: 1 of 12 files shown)


lib/dor/text_extraction/abbyy/file_watcher.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        # These methods control the underlying listener; see:
+        # https://github.com/guard/listen?tab=readme-ov-file#pause--start--stop
+        # NOTE: You need to `sleep` or otherwise avoid letting the process
+        # exit after starting in order for the listener's thread to keep running
+        delegate :start, :pause, :stop, to: :listener
+
+        # Directories where ABBYY places OCR result information
+        RESULT_XML_DIR = 'RESULTXML'
+        EXCEPTIONS_DIR = 'EXCEPTIONS'
```
dnoneill, line 18: These should be set in settings.

lib/dor/text_extraction/abbyy/file_watcher.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        def initialize(root_path:, workflow_updater: nil, listener_options: {})
+          @root_path = root_path
+          @result_xml_path = File.join(root_path, RESULT_XML_DIR)
+          @exceptions_path = File.join(root_path, EXCEPTIONS_DIR)
```
dnoneill, line 23: See above. Entire path should be in settings.

lib/dor/text_extraction/abbyy/file_watcher.rb
thatbudakguy: Didn't realize we had that! Thanks

dnoneill: Actually it's in windows. Might not be good.

thatbudakguy: That might be OK if we set it to the windows version in production/on the VMs though, right? But we can leave it non-windows in local dev?

dnoneill: Well it's only windows on the Abbyy vm, but the common-accessing vm mount is unix and at a different place.

thatbudakguy: Ahh, and we use this setting for two different purposes? Maybe we need two settings then, or something...

lib/dor/text_extraction/abbyy/file_watcher.rb
```diff
@@ -0,0 +1,61 @@
...
+        end
+
+        # Notify SDR that the OCR workflow step completed successfully
+        def process_success(results)
+          @workflow_updater.mark_ocr_completed(results.druid)
+        end
+
+        # Notify SDR that the OCR workflow step failed
+        def process_failure(results)
+          @workflow_updater.mark_ocr_errored(results.druid)
```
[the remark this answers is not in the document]
thatbudakguy: That's a good idea; I'll look for an example of how to do that via the workflow service.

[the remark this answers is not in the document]
thatbudakguy: Added some stuff to report using the error messages that `Results` knows about.

lib/dor/text_extraction/abbyy/file_watcher.rb
thatbudakguy: I split out the two path styles into separate values, and created https://github.com/sul-dlss/shared_configs/pull/2183.

peetucket (approved):

Issue #1220, opened 2024-05-15
Use load balanced preservation catalog API


edsu:
At the moment common-accessioning in production is configured to talk to `preservation-catalog-web-prod-01` but it should talk to `preservation-catalog.stanford.edu` instead. This way we won't overload one node when reading files in nead of OCR from preservation catalog.

The firewall rule to allow access has been done: https://github.com/sul-dlss/operations-tasks/issues/3751

QA and Stage are already pointed at the load balanced preservation-catalog.

PR #1223, opened 2024-05-16
fix typo in reference to pres cat URL setting

jmartin-sul:

noticed this when doing a little poking around and testing for https://github.com/sul-dlss/common-accessioning/issues/1220

(existing) unit tests


edsu (approved):

edsu:
Thanks for catching this @jmartin-sul !

PR #1227, opened 2024-05-16
Raise an error if accession started when version is open

### LINKED ISSUE #1225: Change HB alert to a raise in Start Accession robot
https://github.com/sul-dlss/common-accessioning/blob/main/lib/robots/dor_repo/accession/start_accession.rb#L13

jcoyne:

Fixes #1225 

### DIFF
--- a/lib/robots/dor_repo/accession/start_accession.rb
+++ b/lib/robots/dor_repo/accession/start_accession.rb
@@ -10,7 +10,7 @@ module Robots
         end
 
         def perform_work
-          Honeybadger.notify('[WARNING] Accessioning has been started with an object that is still open') if object_client.version.status.open?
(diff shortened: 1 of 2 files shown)


lib/robots/dor_repo/accession/start_accession.rb
```diff
@@ -10,7 +10,7 @@ def initialize
         end
 
         def perform_work
-          Honeybadger.notify('[WARNING] Accessioning has been started with an object that is still open') if object_client.version.status.open?
+         raise '[WARNING] Accessioning has been started with an object that is still open' if object_client.version.status.open?
```
justinlittman, line 13: Take out the [WARNING]?

justinlittman (approved):
Approve with comment.

PR #1231, opened 2024-05-16
Set up the ABBYY FileWatcher using systemd

thatbudakguy:

Allows running a single persistent service to accomplish #1191.

The systemd service runs only on the primary VM for each load-balanced environment, since only one running instance is needed per ABBYY mount. It is restarted automatically on deploy. The machinery for doing this is largely adapted from searchworks-traject-indexer's `deploy.rb`.

We deployed it to QA and ran the OCR integration test, and the `ocr-create` step completed successfully when the watcher saw that ABBYY placed the output into the right directory.

You can also run the watcher script locally yourself to test it out. First, create and set some sensible paths in a `settings.local.yml` (using your system's root `/tmp` will cause permissions issues):
```yml
sdr:
  abbyy:
    local_result_path: tmp/abbyy/RESULTXML
    local_exception_path: tmp/abbyy/EXCEPTIONS
```
Then invoke it:
```
$ ./bin/abbyy_watcher
I, [2024-05-16T15:52:40.192784 #52380]  INFO -- : Starting...
I, [2024-05-16T15:52:40.192822 #52380]  INFO -- : Watching tmp/abbyy/RESULTXML and tmp/abbyy/EXCEPTIONS for ABBYY result XML files
```
If you add a result XML file from the fixtures into the appropriate directory, you'll see:
```
I, [2024-05-16T15:54:39.939381 #52380]  INFO -- : Found successful OCR results for druid:bb222cc3333 at /Users/budak/Developer/common-accessioning/tmp/abbyy/RESULTXML/bb222cc3333.xml.result.xml
```
...and the watcher will try (and fail) to update the OCR workflow step for that druid.

You can stop the process with CTRL+C.


dnoneill:
Quick question just for my own knowledge. Is there a reason we are having this being run all the time and not just when we know an OCR process is running?

peetucket:
> Quick question just for my own knowledge. Is there a reason we are having this being run all the time and not just when we know an OCR process is running?

I believe it is because the only way we'd know an OCR process is running is when ocrWF is running (which could be n >= 0 at any given time).   So if we kicked off the process from ocrWF, we'd need to ensure it wasn't already running and then we'd need to decide how long it should be running for, since we don't know how long ABBYY will take (might be fast for a small object, or slow for a big object).   

So I guess it's just easier to leave the process always running even though the vast majority of the time it's doing nothing.  As long as there isn't some performance/memory issue, seems ok?  But always wise to consider other options.

justinlittman:
If running the process is a bunch of infra complexity / headache, you just set the ocrWF to be a single process of a single thread. That guarantees that only one will be running at a time.

peetucket:
> If running the process is a bunch of infra complexity / headache, you just set the ocrWF to be a single process of a single thread. That guarantees that only one will be running at a time.

I think it's less headache to just leave it running in the background instead of intelligently starting and stopping it (even if single threaded) because the remedy is simple.  We have a monitor that looks for it - not running?  Start it up.  But this is presuming there isn't some disadvantage to his approach (like a memory leak, it is prone to dying randomly, performance issues on the box eating up CPU, etc.)

peetucket (changes requested):
Please add a cap task which checks the status of the file watcher process called `status` in the same namespace.

thatbudakguy:
Added; you can run it with `cap qa abbyy_watcher_systemd:status`

peetucket (approved):

PR #1241, opened 2024-05-20
Retry when file not found in Preservation

### LINKED ISSUE #1237: Investigate potential race condition in fetch-files
The fetch-files robot sometimes fails with 404s when running the OCR integration test: https://github.com/sul-dlss/infrastructure-integration-test/pull/731

Speculation is that the files are not yet being returned by the preservation client call, even though we waited for `end-accession` to complete before starting `ocrWF`.  Re-running the step will  eventually get it to succeed.  It may be a coincidence, but it seemed that verifying the files were actually available in preservation in Argo (by clicking the preservation download file link) made it work right away after a retry.

Possible Solution:  Wait for files to be available for a certain amount of time by re-trying the call to Preservation a certain number of times?

edsu:

When Preservation doesn't know about a file keep trying 16 times with
an exponential backoff in between requests. This will sleep a maximum of 36
hours, afterwhich it will give up with an error. This is needed because
some objects may still be moving through preservationWF when ocrWF is
invoked.

Closes #1237

Unit tests / CI

### DIFF
--- a/lib/robots/dor_repo/ocr/fetch_files.rb
+++ b/lib/robots/dor_repo/ocr/fetch_files.rb
@@ -19,11 +19,7 @@ module Robots
           ocrable_filenames.each do |filename|
             path = abbyy_path(filename)
             path.parent.mkpath unless path.parent.directory?
-            path.open('wb') do |file_writer|
-              preservation_client.objects.content(druid:,
-                                                  filepath: filename,
-                                                  on_data: proc { |data, _count| file_writer.write(data) })
-            end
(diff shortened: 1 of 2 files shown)


peetucket (commented):
I think this may be a good use of the `retries` gem, which we use in other codebases.  https://github.com/ooyala/retries  Here is one example: https://github.com/sul-dlss/preservation_robots/blob/main/lib/robots/sdr_repo/preservation_ingest/validate_moab.rb#L39-L44

jmartin-sul (approved):

PR #1245, opened 2024-05-22
add in split xml code

### LINKED ISSUE #1176: Split OCR full document XML into page level XML
Write implementation for `ocrWF:split-ocr-xml` robot (see example code below in comments)

ABBYY will create either page level OCR XML or full document OCR XML, but not both.  We want both so the viewer can offer full text search and hit highlighting by page.  

The solution is to ask ABBYY to produce only full document XML and then split it ourselves into page level.  We tested that this will work and have a ruby script below that was used to do the split for the test.

This will be easier to do when we have actual ABBYY output

### LINKED ISSUE #1248: Refactor fetch-files to use new helper method for ABBYY input folder
Update this: https://github.com/sul-dlss/common-accessioning/blob/main/lib/robots/dor_repo/ocr/fetch_files.rb#L34-L38

Use new method `TextExtraction::Ocr#abbyy_input_path` when #1245 is merged

dnoneill:

Fixes #1176 and fixes #1248


lib/dor/text_extraction/ocr.rb
```diff
@@ -11,6 +11,11 @@ def initialize(params = {})
         @workflow_context = params[:workflow_context]
       end
 
+      def abbyy_output_path
+        bare_druid = @cocina_object.externalIdentifier.split(':')[-1]
```
peetucket, line 15: 1. The attr_reader on line 7 means we can don't need to use the instance variable here and would be more consistent with how's it's done in other parts of the codebases
2. We do this to get the bare_druid in other parts of the codebase, so I'd do it this way, again for consistency:

```
 bare_druid = cocina_object.externalIdentifier.delete_prefix('druid:')
```

lib/dor/text_extraction/ocr.rb
```diff
@@ -11,6 +11,11 @@ def initialize(params = {})
...
+        File.join(Settings.sdr.abbyy.local_output_path, bare_druid)
+      end
+
```
peetucket, line 18: While you are making this change, can you add an analogous  `abbyy_input_path` method?  This will come in handy in the as yet unwritten cleanup robot, and we can then refactor the`fetch-files` robot to make use of it (here:  https://github.com/sul-dlss/common-accessioning/blob/main/lib/robots/dor_repo/ocr/fetch_files.rb#L34-L38)

peetucket (changes requested):

lib/dor/text_extraction/ocr.rb
dnoneill: Updated. I was going with (sort of) consistency for the current workcycle with split so we should probably update these methods.

https://github.com/sul-dlss/common-accessioning/blob/bd7c12805820dbe6c5b75726a2c44569a651a3a6/lib/robots/dor_repo/ocr/fetch_files.rb#L30

https://github.com/sul-dlss/common-accessioning/blob/bd7c12805820dbe6c5b75726a2c44569a651a3a6/lib/dor/text_extraction/abbyy/ticket.rb#L26

lib/dor/text_extraction/ocr.rb
dnoneill: Sounds good. Done. I am assuming you want to do this refactor in a new PR and not this one?

peetucket: Yeah, we can do that in a new PR.  I can ticket it once this is in.

lib/dor/text_extraction/ocr.rb
peetucket: Ugh, we do this all over the place in so many different ways.  It's not that big a deal, just an annoyance to keep repeating ourselves in different ways.

PR #1246, opened 2024-05-22
use dor-workflow-client method to set step to error

peetucket:

We have a specific way to put a step in error via the dor-workflow-client and this sets the error message correctly.

Updated specs

### DIFF
--- a/lib/dor/text_extraction/workflow_updater.rb
+++ b/lib/dor/text_extraction/workflow_updater.rb
@@ -19,8 +19,8 @@ module Dor
       end
 
       # Notify SDR that the OCR workflow step failed
-      def mark_ocr_errored(druid, error_message:)
-        @client.update_status(druid:, workflow:, process: step, status: 'error', note: error_message)
+      def mark_ocr_errored(druid, error_msg:)
+        @client.update_error_status(druid:, workflow:, process: step, error_msg:)
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/dor/text_extraction/workflow_updater.rb
```diff
@@ -19,8 +19,8 @@ def mark_ocr_completed(druid)
       end
 
       # Notify SDR that the OCR workflow step failed
-      def mark_ocr_errored(druid, error_message:)
-        @client.update_status(druid:, workflow:, process: step, status: 'error', note: error_message)
+      def mark_ocr_errored(druid, error_msg:)
+        @client.update_error_status(druid:, workflow:, process: step, error_msg:)
       end
```
peetucket, line 24: use this dor-workflow-client when setting an error

jmartin-sul:
motivation for this change: when clumping on the mysterious problem @edsu ran into where a seemingly errored workflow step wasn't blocking subsequent workflow steps (in `ocrWF`) from running, we tried this change, and it fixed that problem (preventing subsequent steps after the failed one from running).

slack conversation for context:
* https://stanfordlib.slack.com/archives/C06EVDJV45N/p1716388441806129
* https://stanfordlib.slack.com/archives/C06EVDJV45N/p1716328604343179?thread_ts=1716325374.743789&cid=C06EVDJV45N

jmartin-sul (approved):

PR #1249, opened 2024-05-23
add abbyy workspace cleanup robot

### LINKED ISSUE #1212: Clear out empty druid subfolder in ABBYY folders
After the ocrWF runs and ABBYY is done OCRing, there will be an empty druid subfolder in the ABBYY input folder.  After we stage the files to the dor workspace, we can also delete the ocr files in the ABBYY output folders.  We can delete these folders.

1. Create a new step right before `end-ocr` called `ocr-workspace-cleanup`.  This requires changes to workflow-server-rails (https://github.com/sul-dlss/workflow-server-rails/pull/771)
2. New robot in common-accessioning.
3. Robot should cleanup for the input (if empty0 and output druid subfolders as well as the XML ticket path.  Paths to cleanup can be obtained using an existing class, see https://github.com/sul-dlss/common-accessioning/blob/main/lib/dor/text_extraction/ocr.rb#L15-L21
4. Raise an error if input folder is not empty

See notes: https://docs.google.com/document/d/1ADOY6Mr9pwVf2EUr2wb-dt8_KVUBxMSksrrcY5Qqi5o/edit#heading=h.xvsjlhm036c3

peetucket:

Fixes #1212 - clears out empty abbyy input and output folders and XML ticket file

Goes with https://github.com/sul-dlss/workflow-server-rails/pull/771

~~TODO: should we expect the OUTPUT to be empty?  or will have it leftover files that are safe to cleanup?~~ It will clear out non-empty folders

~~NOTE: until we actually complete the implementation of the stage-files robot, this workflow step will just error out each time, as nothing will be emptying the output folder, so we may want to hold off merging it for now.~~

New specs

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -21,9 +21,9 @@ sdr:
     remote_output_path: /tmp
     remote_result_path: /tmp
     # Unix-style paths for the common-accessioning VM, used to manage files
-    local_ticket_path: /tmp
+    local_ticket_path: /tmp/input
     local_result_path: /tmp
-    local_output_path: /tmp
+    local_output_path: /tmp/output
     local_exception_path: /tmp
 
 tech_md_service:
--- a/lib/dor/text_extraction/abbyy/ticket.rb
+++ b/lib/dor/text_extraction/abbyy/ticket.rb
@@ -7,10 +7,10 @@ module Dor
       class Ticket
         attr_reader :filepaths, :ocr_languages, :bare_druid
 
-        def initialize(filepaths:, druid:, ocr_languages:)
+        def initialize(filepaths:, druid:, ocr_languages: [])
           @filepaths = filepaths
           @bare_druid = druid.delete_prefix('druid:')
-          @ocr_languages = ocr_languages || []
+          @ocr_languages = ocr_languages
         end
 
         def write_xml
@@ -36,7 +36,7 @@ module Dor
         end
 
         def language_xml
-          return '' if ocr_languages.empty?
+          return '' if ocr_languages.blank?
 
           language_tags = ocr_languages.map { |lang| "<Language>#{lang}</Language>" }.join("\n")
           "<RecognitionParams RecognitionQuality='Balanced' RecognitionMode='FullPage' UsePullXTextAndRecognizeRest='false' LanguageDetectMode='Always'>
--- a/lib/dor/text_extraction/ocr.rb
+++ b/lib/dor/text_extraction/ocr.rb
@@ -30,8 +30,6 @@ module Dor
 
         return true unless Dir.exist?(abbyy_output_path)
 
-        raise "#{abbyy_output_path} is not empty" unless Dir.empty?(abbyy_output_path)
-
         FileUtils.rm_rf(abbyy_output_path)
 
         FileUtils.rm_f(Abbyy::Ticket.new(filepaths: [], druid: cocina_object.externalIdentifier).file_path) # remove XML ticket file
--- a/lib/dor/text_extraction/ocr.rb
+++ b/lib/dor/text_extraction/ocr.rb
@@ -8,7 +8,7 @@ module Dor
 
       def initialize(params = {})
         @cocina_object = params[:cocina_object]
-        @workflow_context = params[:workflow_context]
+        @workflow_context = params[:workflow_context] || {}
         @bare_druid = cocina_object.externalIdentifier.delete_prefix('druid:')
       end
 
@@ -18,6 +18,20 @@ module Dor
 
       def abbyy_input_path
         File.join(Settings.sdr.abbyy.local_ticket_path, bare_druid)
+      end
+
+      def cleanup
+        if Dir.exist?(abbyy_input_path)
+          raise "#{abbyy_input_path} is not empty" unless Dir.empty?(abbyy_input_path)
+
+          FileUtils.rm_rf(abbyy_input_path)
+        end
+
+        return true unless Dir.exist?(abbyy_output_path)
+
+        raise "#{abbyy_output_path} is not empty" unless Dir.empty?(abbyy_output_path)
+
+        FileUtils.rm_rf(abbyy_output_path)
       end
 
       def possible?
--- a/lib/dor/text_extraction/ocr.rb
+++ b/lib/dor/text_extraction/ocr.rb
@@ -20,6 +20,7 @@ module Dor
         File.join(Settings.sdr.abbyy.local_ticket_path, bare_druid)
       end
 
+      # rubocop:disable Metrics/AbcSize
       def cleanup
         if Dir.exist?(abbyy_input_path)
           raise "#{abbyy_input_path} is not empty" unless Dir.empty?(abbyy_input_path)
@@ -32,7 +33,10 @@ module Dor
         raise "#{abbyy_output_path} is not empty" unless Dir.empty?(abbyy_output_path)
 
         FileUtils.rm_rf(abbyy_output_path)
+
+        FileUtils.rm_f(Abbyy::Ticket.new(filepaths: [], druid: cocina_object.externalIdentifier).file_path) # remove XML ticket file
       end
+      # rubocop:enable Metrics/AbcSize
 
       def possible?
         # only items can be OCR'd
--- a/lib/robots/dor_repo/ocr/fetch_files.rb
+++ b/lib/robots/dor_repo/ocr/fetch_files.rb
@@ -26,11 +26,7 @@ module Robots
         def abbyy_path(filename)
           # NOTE: if files of type "file" were allowed here we would have to
           # deal with file hierarchy (subdirectories)
-          Pathname.new(File.join(abbyy_input_path, filename))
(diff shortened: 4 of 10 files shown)


lib/dor/text_extraction/ocr.rb
```diff
@@ -8,7 +8,7 @@ class Ocr
 
       def initialize(params = {})
         @cocina_object = params[:cocina_object]
-        @workflow_context = params[:workflow_context]
+        @workflow_context = params[:workflow_context] || {}
```
peetucket, line 11: this basically makes workflow_context optional by defaulting to an empty hash since some consumers of this class won't need to pass it in (e.g. just getting folder paths or cleaning up)

config/settings.yml
```diff
@@ -21,9 +21,9 @@ sdr:
     remote_output_path: /tmp
     remote_result_path: /tmp
     # Unix-style paths for the common-accessioning VM, used to manage files
-    local_ticket_path: /tmp
+    local_ticket_path: /tmp/input
     local_result_path: /tmp
-    local_output_path: /tmp
+    local_output_path: /tmp/output
     local_exception_path: /tmp
```
peetucket, line 27: differentiate between input and output folders so the tests for cleanup for be more specific

lib/dor/text_extraction/ocr.rb
```diff
@@ -20,6 +20,20 @@ def abbyy_input_path
...
+          FileUtils.rm_rf(abbyy_input_path)
+        end
+
+        return true unless Dir.exist?(abbyy_output_path)
+
+        raise "#{abbyy_output_path} is not empty" unless Dir.empty?(abbyy_output_path)
+
+        FileUtils.rm_rf(abbyy_output_path)
+      end
+
```
peetucket, line 36: this implementation:

- does not cause errors if the folders don't exist (it will just continue)
- will raise an error if the folder exists but is not empty (this is not expected)
- if the input folder is empty, but the output folder is not empty, it will delete the input folder first and then raise an exception on the output folder not being empty ... it is fine to re-run later and it will cleanup the output folder (since non-existence of a folder is not an issue)

lib/robots/dor_repo/ocr/fetch_files.rb
```diff
@@ -26,19 +26,15 @@ def copy_files
...
+          Pathname.new(File.join(ocr.abbyy_input_path, filename))
         end
 
         def ocrable_filenames
           @ocrable_filenames ||= ocr.required? && ocr.possible? ? ocr.filenames_to_ocr : []
         end
 
         def ocr
-          @ocr = Dor::TextExtraction::Ocr.new(cocina_object:, workflow_context: workflow.context)
+          @ocr ||= Dor::TextExtraction::Ocr.new(cocina_object:, workflow_context: workflow.context)
```
peetucket, line 37: refactor: memoize this like we do in the other methods

lib/robots/dor_repo/ocr/fetch_files.rb
```diff
@@ -26,19 +26,15 @@ def copy_files
         def abbyy_path(filename)
           # NOTE: if files of type "file" were allowed here we would have to
           # deal with file hierarchy (subdirectories)
-          Pathname.new(File.join(abbyy_input_path, filename))
-        end
-
-        def abbyy_input_path
-          @abbyy_input_path ||= Dor::TextExtraction::Ocr.new(cocina_object:, workflow_context: workflow.context).abbyy_input_path
+          Pathname.new(File.join(ocr.abbyy_input_path, filename))
```
peetucket, line 29: refactor: we do not need to initialize the `Ocr` class again, we already have it in a helper method below... which makes it not necessary to have an entire method to get the path

lib/robots/dor_repo/ocr/ocr_create.rb
```diff
@@ -11,8 +11,6 @@ def initialize
 
         # available from LyberCore::Robot: druid, bare_druid, workflow_service, object_client, cocina_object, logger
         def perform_work
-          # create XML ticket for ABBYY in shared mount
-
```
peetucket, line 15: this comment is not correct

lib/robots/dor_repo/ocr/split_ocr_xml.rb
```diff
@@ -11,7 +11,7 @@ def initialize
 
         # available from LyberCore::Robot: druid, bare_druid, workflow_service, object_client, cocina_object, logger
         def perform_work
-          base_output_path = Dor::TextExtraction::Ocr.new(cocina_object:, workflow_context: workflow_service).abbyy_output_path
+          base_output_path = Dor::TextExtraction::Ocr.new(cocina_object:).abbyy_output_path
```
peetucket, line 14: we can skip passing the workflow context if all we need is the output path

spec/lib/dor/text_extraction/abbyy/ticket_spec.rb
```diff
@@ -10,6 +10,8 @@
   let(:ticket_xml) { abbyy.send(:xml) }
   let(:fixture_path) { File.join(File.absolute_path('spec/fixtures/ocr'), "#{bare_druid}_abbyy_ticket.xml") }
 
+  before { FileUtils.mkdir_p(File.dirname(abbyy.file_path)) }
+
```
peetucket, line 14: ensure the tmp output path to write the ticket exists

lib/dor/text_extraction/abbyy/ticket.rb
```diff
@@ -7,10 +7,10 @@ module Abbyy
       class Ticket
         attr_reader :filepaths, :ocr_languages, :bare_druid
 
-        def initialize(filepaths:, druid:, ocr_languages:)
+        def initialize(filepaths:, druid:, ocr_languages: [])
           @filepaths = filepaths
           @bare_druid = druid.delete_prefix('druid:')
-          @ocr_languages = ocr_languages || []
+          @ocr_languages = ocr_languages
         end
```
peetucket, line 14: this change allows `ocr_languages` to be optional and defaults to `[]`

lib/dor/text_extraction/abbyy/ticket.rb
```diff
@@ -36,7 +36,7 @@ def output_file_path
...
 
         def language_xml
-          return '' if ocr_languages.empty?
+          return '' if ocr_languages.blank?
```
peetucket, line 40: `.blank?` handles `nil` or `[]`

lib/dor/text_extraction/ocr.rb
jmartin-sul: two questions:
1. is the ticket in the output path?  if not, is gating the ticket file deletion on the existence of the output path intentional?
2. if the input dir is not empty, do we want to refrain from deleting the output dir?

i guess the overall gist of what i'm wondering is, do we want to condition any deletions on other deletions, or do we just want to clean up everything we can safely clean up independently of the contents of paths that can't yet be removed?

lib/dor/text_extraction/ocr.rb
```diff
@@ -8,7 +8,7 @@ class Ocr
 
       def initialize(params = {})
```
jmartin-sul, line 9: not necessarily in scope for this PR since it's this way already, but i wonder if there's any particular reason we're using a params hash here instead of named params?

jmartin-sul (approved):
some nitpicky questions and comments, but nothing to block merging.  might do a tiny follow-on PR in a bit...

lib/dor/text_extraction/ocr.rb
peetucket: 1. Good idea on tmp paths in settings - I'll put a small PR to differentiate the input/output/results folders in tmp.
2. The XML ticket path is configured to be at the root of the INPUT path named by druid, which is why it it's own delete step (else it wouldn't get cleaned up by the other actions), see https://github.com/sul-dlss/common-accessioning/blob/main/lib/dor/text_extraction/abbyy/ticket.rb#L16-L22 and https://github.com/sul-dlss/shared_configs/blob/common-accessioning-qa/config/settings/production.yml#L27   Deleting the XML ticket is the last thing that happens, but the order of output folder vs ticket deletion is probably not super important.
3. If the input dir exists but is not empty, it will raise an error in this implementation, blocking everything else from happening by putting the step in error.  I guess once we see how this goes, if it turns out the input folder is not always empty and this is totally fine and acceptable, we could always remove the raise and just let the cleanup continue.
4. Not sure what the deal is with the params hash in the method signature, I'll fix in the same PR to use named params.

peetucket: https://github.com/sul-dlss/common-accessioning/pull/1277

jmartin-sul: > If the input dir exists but is not empty, it will raise an error in this implementation, blocking everything else from happening by putting the step in error. I guess once we see how this goes, if it turns out the input folder is not always empty and this is totally fine and acceptable, we could always remove the raise and just let the cleanup continue.

sounds like a reasonable plan

Issue #1238, opened 2024-05-17
Rescue exceptions on file watcher and set ocr-create workflow step to error, HB alert


peetucket:
If the abbyy file watcher class throws an error (like a 404 from the workflow service), it will log this in the output, but never return anything to the workflow service, which just leaves the `ocr-create` step in a running state.

Rescue any exceptions from this process, so that we can notify HB and set the workflow step to an error state, so at least we know something went wrong

We can then re-raise the exception for the log

From slack:
- Disentangle handling for an error relevant to a single druid version's ocrWF from an error for the watcher process overall.  Attempt to mark the relevant workflow step failed for this sort of error, with the expectation that it won't be possible to do that if the error is something like a 404 from WFS (or actually, maybe we can handle that by creating the WF step?  I think we do this for preservationAuditWF, now that i think about it).  But either way, send a Honeybadger alert first thing when the error is detected.
- For issues with the watcher process overall... do some sort of HB alert that isn't specific to a workflow step.

PR #1250, opened 2024-05-24
Error reporting improvements

jmartin-sul:

fixes https://github.com/sul-dlss/common-accessioning/issues/1238

* `abbyy_watcher`: Honeybadger alerts for unexpected exits, more consistent use of exit codes (and a very small var name tweak i couldn't resist)
* `fetch-files`: Honeybadger alert when retries exhausted, slightly more detailed message when raising, slight speed up to existing test of retry behavior
* `ocr-create` / `file_watcher`: Honeybadger alert on failure

adjust `.rubocop.yml` to allow slightly longer but still reasonable specs (because a couple spec changes in this PR violated previous default limit).  re-gen `.rubocop_todo.yml`.

- [x] unit tests
- [x] manually test `bin/abbyy_watcher` changes (used local test instructions from #1231, confirmed that signal number was passed along when stopping the process, confirmed that result XML file was still detected)


jmartin-sul:
i think this covers what's requested in the ticket at a high level, because...

> If the abbyy file watcher class throws an error (like a 404 from the workflow service), it will log this in the output, but never return anything to the workflow service, which just leaves the `ocr-create` step in a running state.
> 

We're now rescuing errors from when the `FileWatcher` is instantiated and run in the  systemd script.  We were already logging when `FileWatcher` discovered a processing failure, but this PR adds a little detail, and also a Honeybadger alert to go alongside that (if the work were being done directly within a robot instead of via handoff, a `raise` would get that logging and reporting for free via the lyber-core base class, but the workflow step work is being executed here by the daemon, so that wrapping machinery from the robot framework isn't there).

Unlike with `preservationAuditWF` reporting, a 404 seems very unlikely here, because the filewatcher wouldn't have anything to find if `ocrWF` hadn't been created and triggered file fetching.  So I didn't handle that possibility in this PR, and would suggest revisiting it later if we actually see that problem arise.  But I'm also happy to revisit in this PR or a follow on if others think we should be more proactive about it now.

> Rescue any exceptions from this process, so that we can notify HB and set the workflow step to an error state, so at least we know something went wrong
> 

Per the amendments to the GH issue and Slack conversation, errors at the systemd process level aren't really about a specific instance of a workflow step for a specific druid version, so there's not really a logical workflow step to set to an error state in that case.  But this PR should do what exception handling it can?  With the understanding (from Slack discussion) that the Listen gem's thread for detecting changes will trap and log many errors instead of bubbling them up to the parent process, and so we may discover things like permission errors on the watched directory if we see things piling up in `ocrWF`, which would nudge us to look at the logs.  One more proactive way to deal with a scenario would be an okcomputer or Nagios check for what seems like excessive queue depth or latency on the `ocrWF` `ocr-create` step.  Feels like that should probably be a separate ticket and PR from this if we want to add that, but of course happy to work that into this PR or a follow on under the same ticket if that feels preferable to others.

> We can then re-raise the exception for the log

Logging, reporting to Honeybadger, and exiting with (hopefully) appropriate codes for process level issues that actually `raise`.  We've also added a Nagios check to make sure the watcher process is running, so that should give us another clue if e.g. the process can't consistently stay running.

> From slack:
> - Disentangle handling for an error relevant to a single druid version's ocrWF from an error for the watcher process overall.  Attempt to mark the relevant workflow step failed for this sort of error, with the expectation that it won't be possible to do that if the error is something like a 404 from WFS (or actually, maybe we can handle that by creating the WF step?  I think we do this for preservationAuditWF, now that i think about it).  But either way, send a Honeybadger alert first thing when the error is detected.
> - For issues with the watcher process overall... do some sort of HB alert that isn't specific to a workflow step.

hopefully accomplished in this PR and explained above?

peetucket:
>  One more proactive way to deal with a scenario would be an okcomputer or Nagios check for what seems like excessive queue depth or latency on the `ocrWF` `ocr-create` step. Feels like that should probably be a separate ticket and PR from this if we want to add that, but of course happy to work that into this PR or a follow on under the same ticket if that feels preferable to others.

This is now here: https://github.com/sul-dlss/workflow-server-rails/pull/774

peetucket (approved):

PR #1251, opened 2024-05-24
Move OCR result files to workspace

### LINKED ISSUE #1177: Re-stage content from OCR output
Write implementation for `ocrWF:stage-files` robot

The new caption and OCR robots will be creating new files that will be added to the content folder being accessioned.  

Create a new robot to re-stage the files that came out of OCR in /dor/workspace

Note: closing the version in the last step of ocrWF will kick off accessioning

edsu:

Move Abbyy result files, and split OCR to staging location.

Fixes #1177

Unit tests, and WIP OCR Integration test + looking at logs to make sure the moves happened.


config/settings/test.yml
```diff
@@ -1,2 +1,16 @@
...
+  abbyy:
+    # Windows-style paths for the ABBYY machine, referenced in XML tickets
+    remote_output_path: E:\AbbyyShare\sdr-ocr-qa\OUTPUT
+    remote_result_path: E:\AbbyyShare\sdr-ocr-qa\RESULTXML
+    # Unix-style paths for the common-accessioning VM, used to manage files
+    local_ticket_path: /tmp/ABBYY/INPUT
+    local_result_path: /tmp/ABBYY/RESULTXML
+    local_output_path: /tmp/ABBYY/OUTPUT
+    local_exception_path: /tmp/ABBYY/EXCEPTIONS
+
```
peetucket, line 15: do we really need to override all of these values for the tests?  they are set in `config/settings.yml` to various tmp values

lib/robots/dor_repo/ocr/stage_files.rb
```diff
@@ -9,9 +9,19 @@ def initialize
           super('ocrWF', 'stage-files')
         end
 
-        # available from LyberCore::Robot: druid, bare_druid, workflow_service, object_client, cocina_object, logger
         def perform_work
-          # Copy OCR files from ABBYY output folder to the workspace
+          workspace_dir = object_client.workspace.create
```
peetucket, line 13: if my PRs for [DSA](https://github.com/sul-dlss/dor-services-app/pull/5042) and [dor-services-client](https://github.com/sul-dlss/dor-services-client/pull/436) are accepted, once you bump dor-services-client, you can do this: 

`workspace_dir = object_client.workspace.create(content: true)` and then skip creating the content sub-directory below (it will already be there)

lib/robots/dor_repo/ocr/stage_files.rb
```diff
@@ -9,9 +9,19 @@ def initialize
...
+          move_files(workspace_dir)
+        end
+
+        private
+
+        def move_files
+          content_dir = File.join(workspace_dir, 'content')
+          FileUtils.mkdir(content_dir)
```
peetucket, line 21: won't need the `mkdir` if the DSA work is accepted

lib/robots/dor_repo/ocr/stage_files.rb
peetucket: new dor-services-client gem released, so you can now bump (14.11.0 is what you need) and update that here

deploying updated dsa to QA for testing purposes (we also need the latest version to go with 14.11.0)

config/settings/test.yml
edsu: Yeah, you're right they aren't needed. I removed them.

lib/dor/text_extraction/abbyy/results.rb
```diff
@@ -12,11 +12,21 @@ class Results
         # ab123cd4567.xml.0001.result.xml
         RESULT_FILE_PATTERN = /(?<druid>\w+)\.xml(?:\.(?<run_index>\d+))?\.result\.xml/
 
-        def initialize(result_path:)
```
edsu, line 15: This new method is used to use a druid to locate the latest job result file and instantiate a `Results` object for it. It's a bit clunky that the logger needs to be passed in, but that seems to be what is done elsewhere?

lib/dor/text_extraction/abbyy/results.rb
```diff
@@ -33,7 +43,7 @@ def output_docs
           {}.tap do |structural|
             output_docs.each do |doc|
               doc_type = doc.xpath('@ExportFormat').text.downcase
-              structural[doc_type] = File.join(doc.xpath('@OutputLocation').text, doc.xpath('FileName').text)
+              structural[doc_type] = local_output_path(doc.xpath('@OutputLocation').text, doc.xpath('FileName').text)
```
edsu, line 46: This is needed to convert the Windows location reported by Abbyy to the local Unix location

lib/dor/text_extraction/abbyy/results.rb
```diff
@@ -42,11 +52,43 @@ def alto_doc
           output_docs['alto']
         end
```
edsu, line 54: This does the work of *moving* the Abbyy results files, including split out OCR files, to another location (which will be the workspace directory).

lib/dor/text_extraction/abbyy/results.rb
```diff
@@ -42,11 +52,43 @@ def alto_doc
...
+
+          paths = output_docs.values + split_ocr_paths
+          paths.each do |path|
+            dest_path = File.join(destination_dir, File.basename(path))
+            @logger.info("moving #{path} to #{dest_path}")
+            FileUtils.mv(path, dest_path)
+          end
+        end
+
         private
```
edsu, line 69: This finds any XML files that appear to be OCR files split out from the big one.

lib/robots/dor_repo/ocr/stage_files.rb
```diff
@@ -9,9 +9,18 @@ def initialize
           super('ocrWF', 'stage-files')
         end
```
edsu, line 11: This context helped ensure the workspace was setup properly for each test. I was using TMP_DIR which worked locally but failed in CI for some reason that I couldn't quite figure out.

spec/fixtures/ocr/bb222cc3333.xml.result.xml
```diff
@@ -55,7 +55,7 @@
       UseFindAndFilterTypographicRaster="false" UseSmartColorityDetection="false"
```
edsu, line 55: These are important changes to get the actual file paths from Abbyy to test that they get mapped to local unix directories.

spec/lib/dor/text_extraction/abbyy/results_spec.rb
```diff
@@ -3,10 +3,26 @@
...
 describe Dor::TextExtraction::Abbyy::Results do
-  subject(:results) { described_class.new(result_path:) }
+  subject(:results) { described_class.new(result_path: results_file_path) }
+
+  include_context 'with abbyy dir'
+  include_context 'with workspace dir'
 
   let(:druid) { 'bb222cc3333' }
-  let(:result_path) { File.join(File.absolute_path('spec/fixtures/ocr'), "#{druid}.xml.result.xml") }
+  let(:results_file_path) { "#{abbyy_result_xml_path}/#{druid}.xml.result.xml" }
```
edsu, line 12: Setting up some fake Abbyy results and making sure they can be moved.

lib/dor/text_extraction/abbyy/results.rb
```diff
@@ -12,11 +12,21 @@ class Results
         # ab123cd4567.xml.0001.result.xml
         RESULT_FILE_PATTERN = /(?<druid>\w+)\.xml(?:\.(?<run_index>\d+))?\.result\.xml/
 
-        def initialize(result_path:)
+        # Find the latest Abbyy results XML file by druid, since there coule be more than one.
+        # @param druid [String] The druid to use to look for results
+        # @return [Results, nil] The Results object or nil, if one wasn't found.
+        def self.find_latest(druid:, logger: nil)
+          bare_druid = DruidTools::Druid.new(druid).id
+          result_files = Dir.glob("#{Settings.sdr.abbyy.local_result_path}/#{bare_druid}*.xml")
```
thatbudakguy, line 20: does `Dir.glob` automatically return the results with the most recent first? if not, you could sort them by `run_index` to get the most recent (highest index)

edsu: Yeah, I only realized that it did because my editor told me (maybe rubcop?). 

> If optional keyword sort is given, its value specifies whether the array is to be sorted; the default is true. Passing value false with that keyword disables sorting (though the underlying file system may already have sorted the array).

I think it's worth a comment explaining it, thanks for the nudge.

edsu: Oh wait, actually you're right their sorted in the ascending order. So this is actually a bug.

edsu: Ok I've updated it to get the last one in the list. Hopefully this won't really become needed in practice since we will be removing the result.xml once the workflow completes.

thatbudakguy (approved):

config/settings/test.yml
```diff
@@ -1,2 +1,9 @@
 workflow:
   url: 'http://workflow.example.com/workflow'
+
+sdr:
+  local_workspace_root: /dor/workspace
+  abbyy:
+    # Windows-style paths for the ABBYY machine, referenced in XML tickets
+    remote_output_path: E:\AbbyyShare\sdr-ocr-qa\OUTPUT
+    remote_result_path: E:\AbbyyShare\sdr-ocr-qa\RESULTXML
```
peetucket, line 9: this is fine if it doesn't affect other tests, but just to note that if you want to get different settings for the context of a particular test, you can always do something like this without having to touch the actual config files:

https://github.com/sul-dlss/common-accessioning/blob/main/spec/lib/preserved_file_uris_spec.rb#L68-L71
https://github.com/sul-dlss/common-accessioning/blob/main/spec/robots/dor_repo/ocr/xml_ticket_create_spec.rb#L25-L29

lib/dor/text_extraction/abbyy/results.rb
peetucket: Yeah, not sure what to do about passing this logger everywhere.  A problem for another day.

peetucket (approved):

PR #1256, opened 2024-05-29
stop ocrWF from starting if there are no OCRable files

### LINKED ISSUE #1255: Prevent ocrWF from starting when no available files for OCRing
Currently, ocrWF can start and open a version if there are no available files to be OCRed in the object.  In these cases, this can lead to an "useless" empty version, with no changes and/or things getting hung up at later steps since the XML ticket will have nothing in it.

The user interface in Argo could disable the workflow button in these cases, but the check for OCRable files can be expensive for large objects (as in, many tens of seconds), making it not viable to run on every argo item detail page.  Instead, we can do a simpler check, such as verifying a valid content type.  This would still let the user start OCR when nothing would actually happen.

Possible adjustments:
- throw an error in the first step of ocrWF -- this allows the user to see there is a problem, thereby preventing a new version
- do nothing (accept the empty version)
- something else?

peetucket:

Fixes #1255 - prevent ocrWF from starting if OCR is not possible to run

New specs

### DIFF
--- a/lib/robots/dor_repo/ocr/fetch_files.rb
+++ b/lib/robots/dor_repo/ocr/fetch_files.rb
@@ -34,7 +34,7 @@ module Robots
         end
 
         def ocrable_filenames
-          @ocrable_filenames ||= ocr.required? && ocr.possible? ? ocr.filenames_to_ocr : []
+          @ocrable_filenames ||= ocr.filenames_to_ocr
         end
 
         def ocr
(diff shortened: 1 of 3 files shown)


lib/robots/dor_repo/ocr/fetch_files.rb
```diff
@@ -34,7 +34,7 @@ def abbyy_input_path
         end
 
         def ocrable_filenames
-          @ocrable_filenames ||= ocr.required? && ocr.possible? ? ocr.filenames_to_ocr : []
+          @ocrable_filenames ||= ocr.filenames_to_ocr
         end
```
peetucket, line 38: we don't need to bother doing all of this since the ocrWF will not start if it's not possible now

edsu (approved):

PR #1258, opened 2024-05-29
Create a script that can be used to recover from ABBYY watcher failures

thatbudakguy:
This adds a script that can be run in the event that the watcher process
goes down for some time while OCR is being processed. It checks ABBYY's
result and exception directories and manually updates the workflow
status for any items recorded there, so that the system can be informed
of changes that happened while the watcher was down.

This was tested by putting success/failure result XML files in the
relevant directories locally and invoking the script.

It's intended to not halt when there is an error communicating with
DSA or the workflow service, because some items may already have had
their status updated, so the calls to update them again will fail.

### DIFF
--- a/bin/abbyy_catchup
+++ b/bin/abbyy_catchup
@@ -1,0 +1,38 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# Manually update the OCR workflow status for all items that ABBYY knows about
+
+# This script is intended to be run in production only, in the event of a failure
+# of the ABBYY watcher background service. It will check the ABBYY filesystem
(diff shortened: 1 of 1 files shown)


peetucket:
This is needed because the watcher is looking for changes only, and when done, anything that changed is skipped, so this picks those up.  Is that right?

thatbudakguy:
Yep, that's the idea. If the watcher was off for some time, it won't "see" the changes for any results that appeared while it was off, so this just attempts to process anything that's still on the filesystem.

peetucket (approved):
Will be a good addition to the DevOpsDocs when we write it up

PR #1262, opened 2024-05-31
skip all steps in the ocrWF if object cannot be OCRed

### LINKED ISSUE #1259: Mark all ocrWF steps skipped if pre-reqs not met
Currently ocrWF will put the first step in error if the pre-reqs for OCRing are not met.  We instead just skip the step and all subsequent steps too, with a note indicating why.

Here is the part of the code where it should go: https://github.com/sul-dlss/common-accessioning/blob/main/lib/robots/dor_repo/ocr/start_ocr.rb#L14 replacing the `raise`

This might make it impossible to re-apply the ocrWF to the same version, but a person can always just no-op version the druid and then have a new version to apply ocrWF to.

How the other steps get skipped is up for investigation.  Could add extra workflow context variable that is checked in each step to decide if a skip is warranted or not.  Or could just set all to skip all immediately in the first step.

peetucket:

Fixes #1259 - mark all ocrWF steps as skipped instead of raising an error if the workflow is started but there is nothing to ocr.  This allows the object to be versioned and fixed if needed instead of putting it in error.

~~HOLD for testing on QA~~

See https://argo-qa.stanford.edu/view/druid:bv915yp9350 for test object on argo-qa that was skipped (google book with no images)

Updated specs
QA

### DIFF
--- a/lib/robots/dor_repo/ocr/start_ocr.rb
+++ b/lib/robots/dor_repo/ocr/start_ocr.rb
@@ -28,7 +28,8 @@ module Robots
           note = 'No files available or invalid object for OCR'
           workflow = workflow_service.workflow(pid: druid, workflow_name:)
           # get all of the incomplete steps, except for ourself (start-ocr), we will set that with the usual ReturnState
-          incomplete_steps = workflow.incomplete_processes.map(&:name).reject { |step| step == 'start-ocr' }
+          # and set them to skipped in inverse order (last one gets skipped first to avoid any later steps running)
+          incomplete_steps = workflow.incomplete_processes.map(&:name).reject { |step| step == 'start-ocr' }.reverse
           incomplete_steps.each { |process| workflow_service.update_status(druid:, workflow: workflow_name, process:, status:, note:) }
           LyberCore::ReturnState.new(status: status.to_sym, note:)
         end
--- a/lib/robots/dor_repo/ocr/start_ocr.rb
+++ b/lib/robots/dor_repo/ocr/start_ocr.rb
@@ -11,9 +11,26 @@ module Robots
 
         def perform_work
           raise 'Object is already open' if object_client.version.status.open?
-          raise 'No files available or invalid object for OCR' unless Dor::TextExtraction::Ocr.new(cocina_object:).possible?
 
-          object_client.version.open(description: 'Start OCR workflow')
+          if Dor::TextExtraction::Ocr.new(cocina_object:).possible?
+            object_client.version.open(description: 'Start OCR workflow')
+          else
+            skip_workflow
+          end
+        end
+
+        private
+
+        # skip entire ocrWF workflow without opening version if OCR is not possible for this object
+        def skip_workflow
+          workflow_name = 'ocrWF'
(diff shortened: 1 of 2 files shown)


spec/robots/dor_repo/ocr/start_ocr_spec.rb
```diff
@@ -17,9 +17,7 @@
   let(:object_client) do
     instance_double(Dor::Services::Client::Object, version: version_client, workspace: workspace_client, find: object)
   end
-  let(:ocr) do
-    instance_double(Dor::TextExtraction::Ocr, possible?: possible)
-  end
+  let(:ocr) { instance_double(Dor::TextExtraction::Ocr, possible?: possible) }
```
peetucket, line 21: just tidy this to one line

spec/robots/dor_repo/ocr/start_ocr_spec.rb
```diff
@@ -39,9 +37,32 @@
...
+    let(:return_status) { perform.status }
 
-    it 'raise an error' do
-      expect { perform }.to raise_error('No files available or invalid object for OCR')
+    before do
+      allow(LyberCore::WorkflowClientFactory).to receive(:build).and_return(workflow_client)
+      allow(workflow_client).to receive(:update_status)
+    end
+
+    it 'sets all steps (except self) in the workflow to waiting and returns status of skipped for self' do
```
jmartin-sul, line 59: is "waiting" possibly a copy/paste typo?  my read is that the other steps are also set to skipped?

jmartin-sul (approved):
it seems like this would make things work more smoothly now (good call on skipping the steps in reverse order), so i'm in favor of merging as-is and doing follow up PRs if either of the questions/suggestions warrants it.

lib/robots/dor_repo/ocr/start_ocr.rb
```diff
@@ -11,9 +11,26 @@ def initialize
...
+        # skip entire ocrWF workflow without opening version if OCR is not possible for this object
+        def skip_workflow
+          workflow_name = 'ocrWF'
+          status = 'skipped'
+          note = 'No files available or invalid object for OCR'
+          workflow = workflow_service.workflow(pid: druid, workflow_name:)
+          # get all of the incomplete steps, except for ourself (start-ocr), we will set that with the usual ReturnState
+          incomplete_steps = workflow.incomplete_processes.map(&:name).reject { |step| step == 'start-ocr' }
+          incomplete_steps.each { |process| workflow_service.update_status(druid:, workflow: workflow_name, process:, status:, note:) }
+          LyberCore::ReturnState.new(status: status.to_sym, note:)
```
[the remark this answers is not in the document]
peetucket: Yes, that's a good idea.  I'll make a ticket for this for future work, referencing this PR.  It would be fairly straightforward to implement if we need it again.

spec/robots/dor_repo/ocr/start_ocr_spec.rb
peetucket: typo, will fix before merge

PR #1264, opened 2024-06-03
Update Cocina structural with Abbyy files

### LINKED ISSUE #1185: Update cocina when OCR complete
part of #1166 
depends on #1177

Write implementation for `ocrWF:update-cocina` robot

We need to update the cocina (in a new robot in ocrWF) when OCRing is complete, and prior to accessioning content.  This is similar to how the cocina structural is generated by pre-assembly in [app/lib/pre_assembly/from_staging_location/structural_builder.rb](https://github.com/sul-dlss/pre-assembly/blob/main/app/lib/pre_assembly/from_staging_location/structural_builder.rb#L41-L52)

To simplify things this workflow step can determine what files need to be added by simply looking at what files have been placed in the workspace directory as part of #1177.

New cocina fields at the file level (in structural) will need to be added based on if accessioned content has already had OCR performed and when OCR is done by our new ABBYY based robots.

See https://docs.google.com/document/d/1ScvlgCI-YhyV2LaaDLTVOWTvuGe4bM1tmgzVbAHTDw8/edit#heading=h.og29aoqammz0 for details on the attributes, values and when set.

The values indicating when a file had OCR performed by ABBYY should presumably be set when the create-ocr robot has completed successfully.  The values indicating when a file had incoming provided OCR is also set in pre-assembly, but this will be set in the cocina elsewhere as part of sul-dlss/pre-assembly#1441

What ABBYY returns:
```
Images of text:
1 ALTO
1 PDF
1 plain-text

PDF:
1 PDF (unless otherwise determined that we need additional files)
```

We need to ensure that when cocina is generated, these files are added with the correct attributes so they are shelved, published and preserved correctly.

Ref May 16 notes: https://docs.google.com/document/d/1ADOY6Mr9pwVf2EUr2wb-dt8_KVUBxMSksrrcY5Qqi5o/edit#heading=h.8u9choka43a

edsu:

We need to update a DRO's structural metadata with the new files so that they get accessioned as part of accessionWF.

Closes #1185

New unit tests and new integration tests on the https://github.com/sul-dlss/infrastructure-integration-test/tree/updated-pre-assembly-spec branch. I also manually submitted some items through Preassembly.


lib/dor/text_extraction/abbyy/results.rb
```diff
@@ -64,6 +64,9 @@ def move_result_files(destination_dir)
             dest_path = File.join(destination_dir, File.basename(path))
             @logger.info("moving #{path} to #{dest_path}")
             FileUtils.mv(path, dest_path)
+            # ensure that the file can be read by anyone so that preservation_robots can read it
+            # during preservationIngestWF transfer-object which runs as user pres
+            FileUtils.chmod('u=wr,g=wr,o=r', dest_path)
           end
```
peetucket, line 70: possibly remove if we can have abbyy/ops set these correctly

lib/dor/text_extraction/abbyy/results.rb
```diff
@@ -73,6 +76,7 @@ def move_result_files(destination_dir)
         # NOTE: the multiple page Abbyy OCR file is NOT accessioned.
         def paths_to_move
           paths = (output_docs.values + split_ocr_paths).uniq
+          @logger.info("found these candidate paths to move: #{paths}")
           paths.reject { |path| path.end_with?("#{druid}.xml") }
```
peetucket, line 80: possibly remove once we verify it all works as expected

lib/dor/text_extraction/cocina_updater.rb
```diff
@@ -0,0 +1,222 @@
...
+
+      private
+
+      # update the cocina with new files from the workspace
+      def update_cocina
+        content_dir.children.sort.each do |file|
+          next if file.basename.to_s.start_with?('.')
+
+          if file.basename.to_s.match(/#{bare_druid}\.(pdf|txt)$/)
+            add_file_to_new_resource(file)
```
peetucket, line 47: I think this logic is true of both documents and image content types (verify with Andrew)?   Either way, perhaps an explanatory comment makes sense.

lib/dor/text_extraction/cocina_updater.rb
```diff
@@ -0,0 +1,222 @@
...
+          else
+            add_file_to_resource(file)
+          end
+        end
+      end
+
+      # Rename the PDF that was generated for an Item of type document
+      # TODO: maybe xml_ticket_create.rb should be adjusted to create it this way?
```
peetucket, line 55: interesting idea - so that would be in here somewhere? https://github.com/sul-dlss/common-accessioning/blob/main/lib/dor/text_extraction/abbyy/ticket.rb#L34-L36

lib/dor/text_extraction/cocina_updater.rb
```diff
@@ -0,0 +1,222 @@
...
+      # @param workspace_dir [String] the path where the new files to add are
+      def self.update(dro:, workspace_dir:)
+        CocinaUpdater.new(dro:, workspace_dir:).update
+      end
+
+      def initialize(dro:, workspace_dir:)
+        @dro = dro
+        @new_resources = []
+        @workspace_dir = Pathname.new(workspace_dir)
+      end
```
peetucket, line 30: I think we can ditch all of the @workspace_dir passing around since we can get it from DruidTools directly in this class.

peetucket (changes requested):

lib/dor/text_extraction/cocina_updater.rb
edsu: Yes, I think so? I didn't want to disturb things too much there since I wanted to get this into a working state first.

lib/dor/text_extraction/cocina_updater.rb
edsu: I did try doing that, but the current interface for CocinaUpdater is:

```
CocinaUpdater.update(dro:, workspace_dir:)
```

So it actually was a bit harder to use DruidTools, and not really worth the effort?

peetucket: I'm taking a quick stab at it right now.  We are the only consumer of the class so we can change that signature too.

lib/dor/text_extraction/cocina_updater.rb
edsu: I added a comment. It's my understanding from the examples in QA that it needed to happen for both images and documents. Maybe when @andrewjbtw gets a chance to to do some end-to-end testing he will notice if this is the desired behavior?

lib/dor/text_extraction/abbyy/results.rb
edsu: I've removed it, thanks!

lib/dor/text_extraction/cocina_updater.rb
peetucket: Agreed.  We can come back to that later as a refactor if time allows once everything is working.

lib/dor/text_extraction/abbyy/results.rb
edsu: Yeah, I don't know if that's possible. Perhaps it could be ticketed as follow up work?

peetucket (approved):

peetucket:
I re-verified this after my change on QA with integration tests.

PR #1277, opened 2024-06-07
adjust settings for tests and adjust method signature

peetucket:

Follow on from comments in #1249 

CI


jmartin-sul (approved):

jmartin-sul:
thanks!

PR #1282, opened 2024-06-07
Ensure we replace files correctly

### LINKED ISSUE #1281: New OCR files are not being accessioned
- Start with an object that already has accessioned OCR files from a previous accessioning run (e.g. full PDF and page level XML files).
- Start ocrWF again via Argo using the "Text extraction" button
- It will run completely, but it looks like the new files are not replacing the existing files.

I discovered this on this object: https://argo-qa.stanford.edu/view/druid:hj614hq2225 after I changed how page level XML is created, and I observed that while different page level XML files were being written to the abbyy output folder and then copied to the workspace, the new page level XML files were not present in the new version of the object nor did they make it to preservation.  Only the new PDF was there.

Go digging around the moabs on disk on `preservation-catalog-web-qa-01` and you will only see the updated PDF between versions, not the page level XML:  

`/services-disk-qa/storage-root-2/sdr2objects/hj/614/hq/2225/hj614hq2225/`

peetucket:

(hopefully) fixes #1281 -- when we generate new OCR files, we need to be sure they are replaced in cocina

Updated specs

System testing:

1. Created an OCR item with integration test.
2. Used Preassembly with a file manifest downloaded from Argo to update one of the TIFF files.
3. Ran Text Extraction from Argo.
4. Confirmed the OCR XML was updated.


lib/dor/text_extraction/cocina_updater.rb
```diff
@@ -62,14 +62,29 @@ def rename_document_pdf
...
+        guard_overwrite(path) if file_in_cocina?(path)
+
+        return unless path.exist? # we might have deleted the file if we are guarding against replacing it
+
+        resource = find_resource(path)
+
+        if resource
+          delete_file_from_resource(path, resource) if file_in_cocina?(path) # if it's in cocina, remove it first, so that we can replace it!
+          add_file_to_resource(path, resource)
         else
```
peetucket, line 75: this is the key change, basically we have to remove the file from cocina and then re-add it if a replacement.... previously for replacement, it wasn't touching cocina at all, thus guaranteeing the new files never replaced the old ones

lib/dor/text_extraction/ocr.rb
```diff
@@ -102,9 +102,9 @@ def allowed_mimetypes
       # TODO: refine list of allowed object types for OCR
       def resource_type_mapping
         {
-          'https://cocina.sul.stanford.edu/models/book' => 'page',
-          'https://cocina.sul.stanford.edu/models/document' => 'document',
-          'https://cocina.sul.stanford.edu/models/image' => 'image'
+          Cocina::Models::ObjectType.book => 'page',
+          Cocina::Models::ObjectType.document => 'document',
+          Cocina::Models::ObjectType.image => 'image'
         }
```
peetucket, line 108: unrelated refactor to remove these strings and use the values from CocinaModels gem

spec/lib/dor/text_extraction/cocina_updater_spec.rb
```diff
@@ -380,7 +388,15 @@
...
+                correctedForAccessibility: corrected_for_accessibility,
+                hasMessageDigests: [{
+                  type: 'md5',
+                  digest: 'bogus_md5'
+                },
+                                    {
+                                      type: 'sha1',
+                                      digest: 'bogus_sha1'
+                                    }]
               }
```
peetucket, line 400: this new mocked data lets us test for the replacement of the existing file below

spec/lib/dor/text_extraction/ocr_spec.rb
```diff
@@ -109,7 +109,7 @@ def build_file(sdr_peserve, filename)
   end
 
   describe '#ocr_files' do
-    let(:cocina_object) { instance_double(Cocina::Models::DRO, externalIdentifier: druid, structural:, type: 'https://cocina.sul.stanford.edu/models/document') }
+    let(:cocina_object) { instance_double(Cocina::Models::DRO, externalIdentifier: druid, structural:, type: Cocina::Models::ObjectType.document) }
```
peetucket, line 113: refactor to replace string with value from cocinamodels gem

spec/lib/dor/text_extraction/cocina_updater_spec.rb
```diff
@@ -415,6 +436,11 @@
 
       it 'did not add a new file to the resource' do
         expect(dro.structural.contains[0].structural.contains.length).to eq 2
+        expect(dro.structural.contains[0].structural.contains[1].label).to eq 'page1.xml' # this is the first page
+        # the externalIdentifier and shas are the same as the original in our mocked cocina above
+        expect(dro.structural.contains[0].structural.contains[1].externalIdentifier).to eq "https://cocina.sul.stanford.edu/file/#{bare_druid}_001"
+        expect(dro.structural.contains[0].structural.contains[1].hasMessageDigests[0].attributes).to eq({ type: 'md5', digest: 'bogus_md5' })
+        expect(dro.structural.contains[0].structural.contains[1].hasMessageDigests[1].attributes).to eq({ type: 'sha1', digest: 'bogus_sha1' })
       end
```
peetucket, line 444: specs verifying we replace the file (or not)

these existing specs would have failed below since the file wasn't getting replaced cocina when it should have been

edsu (approved):

edsu (approved):
I reran text-extraction and can see that the new OCR XML files with the `<Description>` element are getting created and pushed to Stacks and Preservation.

But it looks like the Item level `.pdf` and `.txt` files are getting added each time, to new resources? Perhaps that should be ticketed and addressed separately?

peetucket:
I like the changes you made to the logic too, it makes it clearer what is happening with updates vs adds

lib/dor/text_extraction/cocina_updater.rb
```diff
@@ -62,14 +65,31 @@ def rename_document_pdf
...
+        resource = find_resource(path)
+
+        delete_file_from_resource(path, resource)
+        add_file_to_resource(path, resource)
+      end
+
+      def add_file(path)
+        resource = find_resource_with_stem(path)
+
+        if resource && !path.basename.to_s.match(/^#{bare_druid}\.(pdf|txt)$/)
```
peetucket, line 78: Maybe a code comment here could be useful explaining why we are ignoring #{bare_druid}.pdf and .txt  ?   

Though, actually, given the logic change above, shouldn't this just become replacements now instead of getting added over and over again?  In other words, do we even need that `!path.basename.to_s.match(/^#{bare_druid}\.(pdf|txt)$/)` check anymore?  Shouldn't our code above discover the .pdf is in cocina, and then delete and re-add it?

edsu:
The unit tests pass for me, but I've not seen the file update work properly in QA yet. Initially this was because the `file_manifest.csv` I was using to update the Item I had created did not have a column for `sdrGeneratedText`, so it was getting set to `false` which caused the initial OCR XML not to get overwritten. 

Even with a fix deployed to Argo to include that `sdr_generated_text` column in the `file_manifest.csv` the value for `sdrGeneratedText` appears to be getting set to `false` by Preassembly.

I've left some notes on how to replicate the problem here: https://github.com/sul-dlss/argo/pull/4475#pullrequestreview-2111518346

peetucket:
> The unit tests for file update pass for me, but I've not seen the file update work properly in QA yet.
> 
> Initially this faillure was because the `file_manifest.csv` I was using to update the Item I had created did not have a column for `sdrGeneratedText`, so Preassembly was setting it to `false` which caused the initial OCR XML not to get overwritten.
> 
> Even with a fix deployed to Argo to include that `sdr_generated_text` column in the `file_manifest.csv` the value for `sdrGeneratedText` appears to be still be getting set to `false` by Preassembly. Perhaps there is something missing from Preassembly to set read and use it?
> 
> I've left some notes on how to replicate the problem here: [sul-dlss/argo#4475 (review)](https://github.com/sul-dlss/argo/pull/4475#pullrequestreview-2111518346)



> The unit tests for file update pass for me, but I've not seen the file update work properly in QA yet.
> 
> Initially this faillure was because the `file_manifest.csv` I was using to update the Item I had created did not have a column for `sdrGeneratedText`, so Preassembly was setting it to `false` which caused the initial OCR XML not to get overwritten.
> 
> Even with a fix deployed to Argo to include that `sdr_generated_text` column in the `file_manifest.csv` the value for `sdrGeneratedText` appears to be still be getting set to `false` by Preassembly. Perhaps there is something missing from Preassembly to set read and use it?
> 
> I've left some notes on how to replicate the problem here: [sul-dlss/argo#4475 (review)](https://github.com/sul-dlss/argo/pull/4475#pullrequestreview-2111518346)

I think this may fix this problem if I am understanding it correctly: https://github.com/sul-dlss/pre-assembly/pull/1478

Suggest deploying that branch of pre-assembly and trying again.  I am not sure we ever thought about using file manifests in pre-assembly with OCR type content since file manifest are almost always used for media content.

lib/dor/text_extraction/cocina_updater.rb
edsu: Yes, updates don't care about the filename since like you said they just delete and then add from the existing resource. But when the file is being added for the first time it needs to be added to a new resource, so that's why the check is there. I will add a comment because you're right, it is super cryptic.

edsu:
> I think this may fix this problem if I am understanding it correctly: [sul-dlss/pre-assembly#1478](https://github.com/sul-dlss/pre-assembly/pull/1478)
> 
> Suggest deploying that branch of pre-assembly and trying again. I am not sure we ever thought about using file manifests in pre-assembly with OCR type content since file manifest are almost always used for media content.

Yes, that fixed it!

PR #1287, opened 2024-06-13
assume accessioned when opening version

peetucket:

To avoid race conditions

QA


peetucket:
Not the approach we want to take.

PR #1295, opened 2024-06-17
speed up split alto xml step

### LINKED ISSUE #1293: Figure out why split-xml takes a long time for large objects
See https://argo-qa.stanford.edu/catalog?f%5Btag_ssim%5D%5B%5D=books+%3A+larger

Some of these objects have the split-xml step taking hours, which doesn't seem to make sense. 

in particular https://argo-qa.stanford.edu/view/druid:xn427rt3998

peetucket:

Fixes #1293.  For an example large XML file (140 MB) with 483 pages, time is now down to < 10 seconds on my laptop (run on server was hours, test run on my laptop with previous code was killed after a few minutes when only a few pages were done).

This does several things to improve performance:
1. removes the building up in memory of all the page XMLs into a hash, instead writing them out immediately to disk one by one
2. pulls the various nodes that are needed out of the full document once and caches, instead of doing it for each page
3. pulls the namespaces out from the full document once and caches, instead of doing it for each page

During my testing, #2 and #3 were the big culprits, taking almost ten seconds to do for the large 140 MB XML full object file, so if you do this each time through the loop for each page, it takes a long time.

Also:
1. Updated the specs to match the new method names
2. Added expected output files per page so we can verify the format of the XML generated for each page.
3. ~~Raise an error in the robot if you can't find the full document XML file to split -- I think this would be unexpected, so not sure we want to just let the robot complete without some kind of error~~ Not a correct assumption: PDF/doc will not have this.  Skip the step with a note instead of erroring out.


lib/dor/text_extraction/abbyy/split_alto.rb
```diff
@@ -13,9 +13,12 @@ def initialize(alto_path:)
...
 
         def write_files
-          split_to_files.each do |filename, xml_content|
-            File.write(output_path(filename), xml_content)
+          pages = doc.css('//Page')
+          pages.each_with_index do |page, i|
+            filename = "#{File.basename(page_filenames[i], File.extname(page_filenames[i]))}.xml"
+            File.write(output_path(filename), xml_structure(page).to_xml)
           end
+          true
```
peetucket, line 21: If you do not return true (or something else) here, the method return result is an array of nokogiri documents, which can get very very large.  We do not care about the result, so just return true

lib/dor/text_extraction/abbyy/split_alto.rb
```diff
@@ -24,31 +27,30 @@ def output_path(filename)
...
+              xml.parent << ocr_processing_nodes
+              xml.parent << style_nodes
               xml.parent << page
             end
           end
         end
 
-        def page_filenames
-          @page_filenames ||= doc.css('//sourceImageInformation').children.select(&:element?).map(&:text)
+        def namespaces
```
peetucket, line 40: this diff is weird ... `page_filenames` is the same and is shown below (i added other methods above it)

lib/dor/text_extraction/abbyy/split_alto.rb
```diff
@@ -24,31 +27,30 @@ def output_path(filename)
...
-        # rubocop:enable Metrics/AbcSize
 
-        def split_to_files
-          pages = doc.css('//Page')
-          {}.tap do |file_hash|
-            pages.each_with_index do |page, i|
-              filename = "#{File.basename(page_filenames[i], File.extname(page_filenames[i]))}.xml"
-              file_hash[filename] = xml_structure(page).to_xml
-            end
-          end
```
peetucket, line 51: this method is gone - we aren't going to split in memory into a hash anymore since we don't need to ... we will just iterate over `pages` and write to disk immediately

lib/dor/text_extraction/abbyy/split_alto.rb
```diff
@@ -24,31 +27,30 @@ def output_path(filename)
...
+        def style_nodes
+          @style_nodes ||= doc.css('//Styles')
+        end
+
+        def ocr_processing_nodes
+          @ocr_processing_nodes ||= doc.css('//OCRProcessing')
+        end
```
peetucket, line 50: caches these things in memory to avoid parsing out of the full doc each time

lib/dor/text_extraction/abbyy/split_alto.rb
```diff
@@ -24,31 +27,30 @@ def output_path(filename)
...
               xml.parent << page
             end
           end
         end
 
-        def page_filenames
-          @page_filenames ||= doc.css('//sourceImageInformation').children.select(&:element?).map(&:text)
+        def namespaces
+          @namespaces ||= { 'xsi:schemaLocation': 'http://www.loc.gov/standards/alto/ns-v3# http://www.loc.gov/alto/v3/alto-3-1.xsd' }.merge(doc.collect_namespaces)
         end
```
peetucket, line 42: caches namespaces in memory to avoid parsing out of the full doc each time (turns out this is pretty slow on a large XML document)

lib/robots/dor_repo/ocr/split_ocr_xml.rb
```diff
@@ -13,7 +13,7 @@ def initialize
         def perform_work
           base_output_path = Dor::TextExtraction::Ocr.new(cocina_object:).abbyy_output_path
           alto_path = File.join(base_output_path, "#{bare_druid}.xml")
-          return unless File.exist?(alto_path)
+          return LyberCore::ReturnState.new(status: :skipped, note: 'No full object XML file') unless File.exist?(alto_path)
```
peetucket, line 17: set the step to skipped with a note to make it clear nothing happened (this is for PDF/documents, which will not have full XML file)

lib/dor/text_extraction/abbyy/split_alto.rb
dnoneill: This could theoretically be hard coded. It was in the original script. I had it get it from the doc because it seemed safer but it is relatively low risk to assume the namespaces are always going to be the same.

peetucket: yeah ok ... i think it is fine this way for now ...  during my testing, I noticed that `doc.collection_namespaces` took like 20 seconds for a large XML doc, so doing this once per object is acceptable, doing it once per page was bad

if it turns out we still have performance issues we could hardcode too as you suggest

dnoneill (approved):

Issue #1289, opened 2024-06-13
Investigate race condition in ocrWF


peetucket:
ocrWF:start-ocr errored out a few times indicating the object could not be opened during accessioning

This could be because the last step of accessionWF still considered as running.

This shouldn't occur because of https://github.com/sul-dlss/dor-services-app/blob/main/app/services/workflow_state_service.rb#L50 but we should verify this working as expected and modify if needed to avoid race condition.

Error came from here: https://github.com/sul-dlss/dor-services-app/blob/main/app/services/version_service.rb#L164

Or maybe we a retry in start-ocr around the open call (less desirable)

Retry implemented as it occurred again.  See #1321 

Some more backstory: 

Error coming from https://github.com/sul-dlss/dor-services-client/blob/main/lib/dor/services/client.rb#L43-L55

Still no explanation for why this happens sometimes. This happened for a very large object for which a number of steps ran slowly, but that shouldn't matter I don't think: https://argo-qa.stanford.edu/view/druid:xh838rk3862

This is where the source exception is coming from I believe, basically dor-services-client calls DSA and asks it to open, and then .open calls ensure_openable! first, which throws the exception:

https://github.com/sul-dlss/dor-services-app/blob/main/app/services/version_service.rb#L157-L158

This calls https://github.com/sul-dlss/dor-services-app/blob/main/app/services/workflow_state_service.rb#L55-L59 which looks for the lifecycle of accessioned via the workflow service, which I guess is coming back as missing at first. Since all of the calls are to workflow-server-rails, once a completed step of end-accession is there, it should come back.

This looks potentially suspicious, but has to do with solr which I don't think should matter in our case: https://github.com/sul-dlss/workflow-server-rails/blob/main/app/controllers/steps_controller.rb#L57-L61

peetucket:
It is possible this is related to the issues fixed in https://github.com/sul-dlss/common-accessioning/pull/1290 but not 100% sure.  May want to get that in first and see if this recurs before spending a lot of time chasing this down.

peetucket:
Potentially add retry but with extra logging to see when this happens?

peetucket:
Not having seen this happen recently, closing until it actually happens again.

peetucket:
Analysis:

* ocrWF:start-ocr can be started by accessionWF:end-accession under certain conditions.  See https://github.com/sul-dlss/common-accessioning/blob/main/lib/robots/dor_repo/accession/end_accession.rb#L25-L36
[...]

PR #1302, opened 2024-06-19
add retries and extra logging for failure to open object

peetucket:

As a way to investigate/mitigate #1289 ?

QA

### DIFF
--- a/lib/robots/dor_repo/ocr/start_ocr.rb
+++ b/lib/robots/dor_repo/ocr/start_ocr.rb
@@ -30,7 +30,7 @@ module Robots
             tries += 1
             raise e unless tries < 3
 
-            Honeybadger.notify("[NOTE] Problem opening object version at start of ocrWF for #{druid}", context: { druid:, tries:, error: e })
+            Honeybadger.notify('[NOTE] Problem opening object version at start of ocrWF', context: { druid:, tries:, error: e })
             retry
           end
         end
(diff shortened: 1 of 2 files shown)


lib/robots/dor_repo/ocr/start_ocr.rb
```diff
@@ -11,14 +11,29 @@ def initialize
...
+
+        def open_object
+          tries = 0
+          begin
+            object_client.version.open(description: 'Start OCR workflow')
+          rescue Dor::Services::Client::Error => e
+            tries += 1
+            raise e unless tries < 3
+
+            Honeybadger.notify("[NOTE] Problem opening object version at start of ocrWF for #{druid}", context: { druid:, tries:, error: e })
```
jmartin-sul, line 33: i think we might get slightly better notifications if the `for #{druid}` is removed from the notification text and left only in the `context` hash, since that should allow for easier error message collation by honeybadger.  i think it might still do the right thing, but it might also lead to totally separate notifications for each druid, instead of one encompassing notification bucket with a bunch of individual instances.

jmartin-sul (approved):

jmartin-sul (approved):
one small suggestion.  feel free to ignore, fix and merge, or merge and fix in a follow-on.

lib/robots/dor_repo/ocr/start_ocr.rb
peetucket: good call, just removed druid from the text of the alert

jmartin-sul (approved):

jmartin-sul:
ah, rubocop is probably complaining because you can use single quotes for that string now that you're not doing any variable interpolation.  but LGTM, thanks for revisiting, feel free to merge once rubocop has been appeased.

jmartin-sul (approved):
thrice approved, approved yet again, and thusly merged

PR #1303, opened 2024-06-20
add some delay in retry

peetucket:

Follow on to #1302 to better match what is in fetch-files ... if we hit an error in opening the version, add a bit of delay before retrying again.

CI


justinlittman:
Do you have a theory as to what is causing this problem?

peetucket:
> Do you have a theory as to what is causing this problem?

No, not yet.  We are hoping to get some clues from the HB exception logging if/when it happens again.  If it never happens once in production, we can remove the retry stuff.

dnoneill (approved):

PR #1304, opened 2024-06-20
reduce nokogiri memory consumption

peetucket:

Nokogiri consumption is very large for big XML docs (240 MB doc --> 7 GB in memory nokogiri doc).  This reduces it by about 60% when in memory...though we continue to fill up memory by parsing pages, this allows this ~1100 page document to complete using less than than 8 GB of memory available on QA: https://argo-qa.stanford.edu/view/druid:gq110sz4835

Server and laptop


justinlittman (approved):

peetucket:
Unfortunately after some more testing, it didn't make a big difference...so working on this some more.

justinlittman:
Do you need to use Nokogiri's sax parser?

spec/fixtures/ocr/expected_bb222cc3333_00_0001.xml
```diff
@@ -26,7 +17,7 @@
 <ParagraphStyle ID="StyleId-44B0F6CE-B77C-4378-B4F7-1CCA36C002B3-" ALIGN="Left" LEFT="0." RIGHT="0." FIRSTLINE="0."/>
 <ParagraphStyle ID="StyleId-D40750C1-ED42-4B5E-A658-06B58631CB5D-" ALIGN="Left" LEFT="0." RIGHT="0." FIRSTLINE="0."/>
 </Styles>
-  <Page ID="Page1" PHYSICAL_IMG_NR="1" HEIGHT="10698" WIDTH="6927">
+<Page xmlns="http://www.loc.gov/standards/alto/ns-v3#" ID="Page1" PHYSICAL_IMG_NR="1" HEIGHT="10698" WIDTH="6927">
```
peetucket, line 20: the only real change to the structure of the XML (other than some carriage returns) is the addition of this `xmlns="http://www.loc.gov/standards/alto/ns-v3#"` to the nodes we add in (`Page`, `OcrProcessing` and `Styles`).   I couldn't find a simple way to ditch and I'm hoping it won't matter

peetucket: Seems to be fine, see https://sul-purl-stage.stanford.edu/zd178rm8212

edsu: Nice! The default namespace for the document is `http://www.loc.gov/standards/alto/ns-v3#"` so it doesn't need to be repeated.

lib/dor/text_extraction/abbyy/split_alto.rb
```diff
@@ -27,30 +23,46 @@ def output_path(filename)
...
+          "<?xml version=\"1.0\"?>\n<alto #{namespaces}>\n#{ocr_processing_nodes}\n#{style_nodes}\n#{page}\n</alto>\n"
+        end
+
+        # NOTE: do not use Nokogiri to load and parse the XML in memory at once as it is memory intensive
+        # first pass through the XML to get the common nodes
+        def fetch_common_nodes
+          Nokogiri::XML::Reader(File.open(alto_path)).each do |node|
+            @style_nodes = node.outer_xml if node?(node, 'Styles')
+            @ocr_processing_nodes = node.outer_xml if node?(node, 'OCRProcessing')
+            @page_filename_nodes = node.outer_xml if node?(node, 'sourceImageInformation')
```
edsu, line 37: I didn't know about `outer_xml()`, that's a good one to remember.

edsu (approved):

lib/dor/text_extraction/abbyy/split_alto.rb
peetucket: yeah, useful here

peetucket:
The repeated namespace URI seemed to be added by the nokogiri reader when pulling nodes out via `outer_html` and there wasn't an obvious way to remove it.  It's definitely not needed but doesn't seem to do any harm based on my test.

PR #1309, opened 2024-06-21
add extra logging around deletes/move XML result cleanup

peetucket:

1. Trying to learn more about why files/folders do not seem to get deleted sometimes.
2. Code normalization for consistency.
3. Move Result XML file cleanup to the cleanup step when the file is no longer needed (reverting from #1311)

QA and rspec

### DIFF
--- a/lib/dor/text_extraction/abbyy/file_watcher.rb
+++ b/lib/dor/text_extraction/abbyy/file_watcher.rb
@@ -52,7 +52,7 @@ module Dor
         # Controlled by the `start`, `pause`, and `stop` methods
         def listener
           @listener ||= Listen.to(result_xml_path, exceptions_path, **@listener_options) do |_modified, added, _removed|
-            results = added.map { |path| Dor::TextExtraction::Abbyy::Results.new(result_path: path, logger: @logger) }
+            results = added.map { |path| Dor::TextExtraction::Abbyy::Results.new(result_path: path, logger:) }
             successes, failures = results.partition(&:success?)
             successes.each(&method(:process_success))
             failures.each(&method(:process_failure))
--- a/lib/dor/text_extraction/abbyy/file_watcher.rb
+++ b/lib/dor/text_extraction/abbyy/file_watcher.rb
@@ -67,7 +67,6 @@ module Dor
         end
 
         # Notify SDR that the OCR workflow step failed
-        # rubocop:disable Metrics/AbcSize
         def process_failure(results)
           logger.info "Found failed OCR results for druid:#{results.druid} at #{results.result_path}: #{results.failure_messages.join('; ')}"
           context = { druid: "druid:#{results.druid}", result_path: results.result_path, failure_messages: results.failure_messages }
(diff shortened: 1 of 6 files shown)


lib/dor/text_extraction/abbyy/file_watcher.rb
```diff
@@ -11,7 +11,7 @@ module Abbyy
       # See bin/abbyy_watcher for the script that starts this class
       # See lib/capistrano/tasks/abbyy_watcher_systemd.cap for the systemd service definition
       class FileWatcher
-        attr_reader :result_xml_path, :exceptions_path
+        attr_reader :result_xml_path, :exceptions_path, :logger, :workflow_updater
```
peetucket, line 15: we tend to use attr_reader instead of instance variables throughout method calls with classes

lib/dor/text_extraction/ocr.rb
```diff
@@ -20,20 +21,29 @@ def abbyy_input_path
...
         if Dir.exist?(abbyy_input_path)
           raise "#{abbyy_input_path} is not empty" unless Dir.empty?(abbyy_input_path)
 
+          # e.g. /abbyy/INPUT/ab123cd4567
+          logger.info "Removing empty ABBYY input directory: #{abbyy_input_path}"
           FileUtils.rm_rf(abbyy_input_path)
         end
 
-        return true unless Dir.exist?(abbyy_output_path)
-
```
peetucket, line 32: do not need this line since `FileUtils.rm_rf` doesn't fail if the folder isn't there

lib/dor/text_extraction/ocr.rb
```diff
@@ -20,20 +21,36 @@ def abbyy_input_path
...
+        logger.info "Removing XML Ticket File: #{xml_ticket_file}"
+        FileUtils.rm_f(xml_ticket_file)
+
+        # e.g. /abbyy/RESULTXML/ab123cd4567.xml.result.xml
+        abbyy_results = Dor::TextExtraction::Abbyy::Results.find_latest(druid: bare_druid, logger:)
+        if abbyy_results # this could be nil if there is no latest result XML file located
+          logger.info "Removing XML Result File: #{abbyy_results.result_path}"
+          FileUtils.rm_f(abbyy_results.result_path)
+        end
+
```
peetucket, line 50: delete the result XML file here instead of in the file watcher

lib/robots/dor_repo/ocr/stage_files.rb
```diff
@@ -20,6 +20,8 @@ def move_files(workspace_dir)
           content_dir = File.join(workspace_dir, 'content')
 
           abbyy_results = Dor::TextExtraction::Abbyy::Results.find_latest(druid:, logger:)
+          raise 'No ABBYY Result XML file found' unless abbyy_results
+
```
peetucket, line 24: this shouldn't happen (i.e. the result XML file should be still be there) but if for some reason none are found, this gives us a better error message than what would happen now (which is this: `NoMethodError: undefined method 'move_result_files' for nil abbyy_results.move_result_files(content_dir)` (https://app.honeybadger.io/projects/52894/faults/108106645/01J0YEE9S295KVY5ZTYWAQ3CMJ?page=0)

peetucket:
Output folder sometimes doesn't get removed, and when I tried deleting one on the server it complained "it wasn't empty" even though it looked empty and even though I was using `rm -rf` ... and then after a few tries, it was deleted.  So possibly a weird file system issue.

edsu (approved):

PR #1316, opened 2024-06-24
cleanup exception XML files

### LINKED ISSUE #1314: Delete exception files in abbyy mount during cleanup
I think we want to add deletion of the exception files (in /abbyy/EXCEPTIONS) associated with a druid in the ocr-cleanup step, otherwise they may get picked up again by the abbyy_catchup script.

peetucket:

Fixes #1314 - deletes the exception XML files after ocr-cleanup runs (presumably we have fixed errors if we reached this step).  This allows the abbyy_catchup script to be re-runnable so it won't reprocess these xml files.

Note that it won't cleanup images in the folder -- we don't know exactly which files go with which druids without more work.  The general cleanup rake task in #1313 can get to these later.

New spec

### DIFF
--- a/lib/dor/text_extraction/ocr.rb
+++ b/lib/dor/text_extraction/ocr.rb
@@ -48,6 +48,13 @@ module Dor
           FileUtils.rm_f(abbyy_results.result_path)
         end
 
+        # e.g. /abbyy/EXCEPTIONS/druid:ab123cd4567.xml.result.xml
+        abbyy_exceptions = Dir.glob("#{Settings.sdr.abbyy.local_exception_path}/*#{bare_druid}*")
+        abbyy_exceptions.each do |abbyy_exception_file|
+          logger.info "Removing XML Exception File: #{abbyy_exception_file}"
+          FileUtils.rm_f(abbyy_exception_file)
+        end
+
         true
       end
       # rubocop:enable Metrics/MethodLength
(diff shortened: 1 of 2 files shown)


lib/dor/text_extraction/ocr.rb
```diff
@@ -48,6 +48,13 @@ def cleanup
           FileUtils.rm_f(abbyy_results.result_path)
         end
 
+        # e.g. /abbyy/EXCEPTIONS/druid:ab123cd4567.xml.result.xml
+        abbyy_exceptions = Dir.glob("#{Settings.sdr.abbyy.local_exception_path}/*#{bare_druid}*")
```
jmartin-sul, line 52: idk if we expect this dir to only contain XML tickets for exceptions.  but either way, maybe this glob should include the `.xml` extension (so e.g. `*#{bare_druid}*.xml`), just as an extra safeguard?

jmartin-sul (approved):

jmartin-sul (approved):
i'm going to go ahead and merge this, and make my nitpicky suggestion part of a miscellaneous touchup PR i'm about to put up...

lib/dor/text_extraction/ocr.rb
peetucket: yes good call...

there will possibly be image files that are bad also here, those will remain (though the general cleanup rake task will remove anything older than 1 month if run)

PR #1319, opened 2024-06-25
Retry cleanup

### LINKED ISSUE #1315: Ouput folder not always deleted by ocr cleanup class/robot
Seems reproducible, not sure why it is happening.   Possibly odd mountiung issues/timing.

1. Run integration test.
2. It succeeds.
3. Look at log and see that the logging indicates the output folder was deleted.  No errors.
4. Go to the file system, it's still there.
5. Try and delete it again, and it works.

e.g. After integration test passes on the console, look at logs:

```
I, [2024-06-24T13:10:55.744967 #3137688]  INFO -- : Removing empty ABBYY input directory: /abbyy/INPUT/zz689fz1396
I, [2024-06-24T13:10:55.746523 #3137688]  INFO -- : Removing ABBYY output directory: /abbyy/OUTPUT/zz689fz1396
I, [2024-06-24T13:10:55.749515 #3137688]  INFO -- : Removing XML Ticket File: /abbyy/INPUT/zz689fz1396.xml
I, [2024-06-24T13:10:55.755287 #3137688]  INFO -- : Removing XML Result File: /abbyy/RESULTXML/zz689fz1396.xml.result.
```

Then go to the console:
```
dir='/abbyy/OUTPUT/zz689fz1396'
Dir.exist?(dir)
=> true
FileUtils.rm_rf(dir)
=> ["/abbyy/OUTPUT/zz689fz1396"]
Dir.exist?(dir)
=> false
```

peetucket:

Fixes #1315 

There may be an issue with the file system sometimes throwing an exception when we try to delete a folder, only for it to succeed on a later attempt (I've seen this happen with a couple times, where the robot completed successfully, but a folder was still there, I tried to delete it on the terminal, it failed, I tried again, and it succeeded).  

Our use of `FileUtils.rm_rf` swallows any errors, so if an error does occur, we won't even know anything has happened and instead the folder will remain.  See https://github.com/ruby/fileutils/issues/57   This PR replaces that call with `FileUtils.rm_r`, which also deletes (non-empty) folders, but WILL thrown an exception if a problem occurs.  This lets us at least catch it and retry.

Note that we now need to verify that the folder/file exists before trying to delete it, because it will now throw an exception for a missing file/folder (which we actually don't care about, it's just a noop).

New specs

### DIFF
--- a/lib/dor/text_extraction/ocr.rb
+++ b/lib/dor/text_extraction/ocr.rb
@@ -3,6 +3,7 @@
 module Dor
   module TextExtraction
     # Determine if OCR is required and possible for a given object
+    # rubocop:disable Metrics/ClassLength
     class Ocr
       attr_reader :cocina_object, :workflow_context, :bare_druid, :logger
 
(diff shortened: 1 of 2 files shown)


lib/dor/text_extraction/ocr.rb
```diff
@@ -89,6 +77,52 @@ def filenames_to_ocr
...
+      end
+
+      # e.g. /abbyy/EXCEPTIONS/druid:ab123cd4567.xml.result.xml
+      def cleanup_abbyy_exceptions
+        abbyy_exceptions = Dir.glob("#{Settings.sdr.abbyy.local_exception_path}/*#{bare_druid}*.xml")
+        abbyy_exceptions.each do |abbyy_exception_file|
+          logger.info "Removing XML Exception File: #{abbyy_exception_file}"
+          FileUtils.rm_r(abbyy_exception_file)
+        end
+      end
```
peetucket, line 124: move all of various cleanup bits into their own methods to make the `cleanup` method itself easier to reason with and easily retriable

also guard against the files/folders not existing ... will just be a noop

edsu (approved):
Looks good. 

Aside: what can we rely on if we can't rely on the filesystem?! 💫

peetucket:
> Looks good.
> 
> Aside: what can we rely on if we can't rely on the filesystem?! 💫

yeah i am annoyed we have to do this.... i guess we will see from the HB alerts how often it happens, maybe after all of the file system churn is done, it will work as expected

PR #1320, opened 2024-06-26
Make sure controlmaster is setup and one-time key is generated

mjgiarlo:
Did not test, but you may need these explicit prerequisites until we come up with a better solution in dlss-capistrano. Recommend testing before merging!

### DIFF
--- a/lib/capistrano/tasks/abbyy_watcher_systemd.cap
+++ b/lib/capistrano/tasks/abbyy_watcher_systemd.cap
@@ -2,7 +2,7 @@
 
 namespace :abbyy_watcher_systemd do
   desc 'Restart the abbyy_watcher systemd service'
-  task :restart do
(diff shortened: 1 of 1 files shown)


peetucket (approved):
cool - this worked for me (status check at least)

only downside is that it also applies to things like restarting sidekiq, etc. that we probably have sprinkled through the docs

i wonder if there is a way to do this is a default for all cap tasks?  or maybe that's bad?

mjgiarlo:
@peetucket 💬 
> cool - this worked for me (status check at least)

:shipit: 
 
> only downside is that it also applies to things like restarting sidekiq, etc. that we probably have sprinkled through the docs
> 
> i wonder if there is a way to do this is a default for all cap tasks? or maybe that's bad?

This is what I've started looking into.

PR #1338, opened 2024-09-03
do not HB alert on each retry; reduce parsing of XML

peetucket:

Does five things:

1. Do not send an HB alert each time we attempt to retry deletion of a folder.  This better mimics what is done elsewhere in the code here https://github.com/sul-dlss/common-accessioning/blob/main/lib/robots/dor_repo/ocr/fetch_files.rb#L47-L60 where we only alert if there is an eventual failure.
2. Do not parse the result XML file if we only need to see if a file exists with `find_latest`, otherwise the ocr-workspace-cleanup step is going to have to  unnecessary re-parse that XML file when it all wants to do is delete the file.
3. Stop this HB alert from triggering for PDF docs (which do not have an ALTO file to parse): https://app.honeybadger.io/projects/52894/faults/111166939
4. Adds an extra retry to the cleanup (was still seeing some flakiness even on main)
5. Adds some extra logging to the cleanup to see if we get any more clues as to why it's needed.

Unit tests and on stage

### DIFF
--- a/lib/dor/text_extraction/abbyy/results.rb
+++ b/lib/dor/text_extraction/abbyy/results.rb
@@ -127,7 +127,7 @@ module Dor
         # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         # Software and version used to do OCR processing, found in ALTO output
         def processing_metadata
-          return {} unless File.exist? alto_doc # PDFs don't have ALTO, so just bail out
+          return {} unless alto_doc && File.exist?(alto_doc) # PDFs don't have ALTO, so just bail out
 
           Nokogiri::XML::Reader(File.read(alto_doc)).each do |node|
             next unless node.name == 'processingSoftware'
@@ -140,7 +140,7 @@ module Dor
             break
           end
         rescue StandardError # If we can't parse the ALTO, log it and move on
-          Honeybadger.notify('Failed to parse processing metadata from ALTO XML', context: { alto_doc: })
+          Honeybadger.notify('Failed to parse processing metadata from ALTO XML', context: { result_path:, alto_doc: })
           {}
         end
         # rubocop:enable Metrics/AbcSize, Metrics/MethodLength
--- a/lib/dor/text_extraction/abbyy/results.rb
+++ b/lib/dor/text_extraction/abbyy/results.rb
@@ -14,13 +14,11 @@ module Dor
 
         # Find the latest Abbyy results XML file by druid, since there could be more than one.
         # @param druid [String] The druid to use to look for results
-        # @param logger [Logger] An optional logger to use
-        # @return [Results, nil] The Results object or nil, if one wasn't found.
-        def self.find_latest(druid:, logger: nil)
+        # @return [xml_filepath, nil] The result xml file path or nil, if one wasn't found.
+        def self.find_latest(druid:)
           bare_druid = DruidTools::Druid.new(druid).id
           # NOTE: Dir.glob will sort the filenames in ascending order and we want the last one.
-          result_files = Dir.glob("#{Settings.sdr.abbyy.local_result_path}/#{bare_druid}*.xml")
-          Results.new(result_path: result_files.last, logger:) unless result_files.empty?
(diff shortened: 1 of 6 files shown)


lib/dor/text_extraction/abbyy/results.rb
```diff
@@ -129,7 +127,7 @@ def local_output_dir(windows_dir)
         # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         # Software and version used to do OCR processing, found in ALTO output
         def processing_metadata
-          return {} unless File.exist? alto_doc # PDFs don't have ALTO, so just bail out
+          return {} unless alto_doc && File.exist?(alto_doc) # PDFs don't have ALTO, so just bail out
```
peetucket, line 130: this change will stop this HB alert from being triggered needlessly: https://app.honeybadger.io/projects/52894/faults/111166939

lib/dor/text_extraction/abbyy/results.rb
```diff
@@ -14,13 +14,11 @@ class Results
...
-        # @param logger [Logger] An optional logger to use
-        # @return [Results, nil] The Results object or nil, if one wasn't found.
-        def self.find_latest(druid:, logger: nil)
+        # @return [xml_filepath, nil] The result xml file path or nil, if one wasn't found.
+        def self.find_latest(druid:)
           bare_druid = DruidTools::Druid.new(druid).id
           # NOTE: Dir.glob will sort the filenames in ascending order and we want the last one.
-          result_files = Dir.glob("#{Settings.sdr.abbyy.local_result_path}/#{bare_druid}*.xml")
-          Results.new(result_path: result_files.last, logger:) unless result_files.empty?
+          Dir.glob("#{Settings.sdr.abbyy.local_result_path}/#{bare_druid}*.xml").last
```
peetucket, line 21: this method now just finds the file, it does not initialize the result object and thus try to parse the XML

lib/dor/text_extraction/ocr.rb
```diff
@@ -38,15 +38,13 @@ def cleanup
           tries += 1
           sleep(3**tries)
 
-          raise e unless tries < 4
+          retry if tries < max_tries
 
-          Honeybadger.notify('[NOTE] Problem deleting files and folders in ocrWF:ocr-workspace-cleanup', context: { druid: bare_druid, tries:, error: e })
-          retry
+          raise e
```
peetucket, line 43: do not send HB alerts if the output folder retried, we will still get an HB alert if it actually fails after multiple tries

lib/dor/text_extraction/ocr.rb
```diff
@@ -105,13 +103,13 @@ def cleanup_xml_ticket
...
+        result_path = Dor::TextExtraction::Abbyy::Results.find_latest(druid: bare_druid)
 
         # this could be nil if there is no latest result XML file for this druid
-        return unless abbyy_results && File.exist?(abbyy_results.result_path)
+        return unless result_path && File.exist?(result_path)
 
-        logger.info "Removing XML Result File: #{abbyy_results.result_path}"
-        FileUtils.rm_r(abbyy_results.result_path)
+        logger.info "Removing XML Result File: #{result_path}"
+        FileUtils.rm_r(result_path)
```
peetucket, line 112: changes needed given new logic of `find_latest` method

lib/robots/dor_repo/ocr/stage_files.rb
```diff
@@ -19,9 +19,10 @@ def perform_work
         def move_files(workspace_dir)
           content_dir = File.join(workspace_dir, 'content')
 
-          abbyy_results = Dor::TextExtraction::Abbyy::Results.find_latest(druid:, logger:)
-          raise 'No ABBYY Result XML file found' unless abbyy_results
+          result_path = Dor::TextExtraction::Abbyy::Results.find_latest(druid:)
+          raise 'No ABBYY Result XML file found' unless result_path
 
+          abbyy_results = Dor::TextExtraction::Abbyy::Results.new(result_path:, logger:)
```
peetucket, line 25: changes needed given new logic of `find_latest` method

mjgiarlo (approved):

PR #1339, opened 2024-09-07
another attempt to handle folder deletion

peetucket:

See if this alternate way of deleting folders makes any difference to the flakiness of empty/non-empty folder deletion.

Pending


peetucket:
conclusion: no difference (unless you add the force: true) setting, which then doesn't throw an exception but then also doesn't delete the directory

PR #1350, opened 2024-09-21
Add the ability to send files to an AWS S3 bucket from fetch-files robot

### LINKED ISSUE #1352: Investigate/add the ability to stream files to S3 from preservation
See sul-dlss/common-accessioning#1345

We will be likely fetching files from preservation and send directly to S3 (the analogy from OCR is fetching from preservation and storing to local mount).

Can we stream directly to S3 from preservation in the same way so we can do this in one robot step?  Let's find out.

peetucket:

Fixes #1352 and #1355
 
* Initial implementation of sending files to an S3 bucket from the new `FileFetcher` class.
* Refactors `FileFetcher` method as described in https://github.com/sul-dlss/common-accessioning/pull/1349/files/21fc07f6f19c9b5a7bfc3669f4d7522891b1db5e#r1771797902

~~Depends on #1349~~

Assuming you have configured your local codebase with the correct AWS and Preservation client keys:

```
bin/console

aws_provider ||= Dor::TextExtraction::AwsProvider.new(region: Settings.aws.region, access_key_id: Settings.aws.access_key_id, secret_access_key: Settings.aws.secret_access_key)

druid='druid:vn995rb5304' # actual druid on stage
bare_druid='vn995rb5304'
filename='testocr.xml' # actual file in this object
file_fetcher ||= Dor::TextExtraction::FileFetcher.new(druid:, logger: Logger.new('test.log'))
file_fetcher.write_file_with_retries(filename:,location:'tmp/test.xml') # will write it to your local disk
=> true

file_fetcher.write_file_with_retries(filename:, location: aws_provider.bucket.object(File.join(bare_druid, filename))) # will send it to AWS
=> true

results = aws_provider.bucket.objects
results.map(&:key)
=> ["vn995rb5304/testocr.xml"]
```

- [x] Updated specs.
- [x] Deploy and run OCR integration tests.
- [x] Need to verify it works with actual S3 credentials.


lib/dor/text_extraction/file_fetcher.rb
```diff
@@ -80,6 +79,14 @@ def fetch_and_write_file_to_disk(filename:, path:)
       def preservation_client
         @preservation_client ||= Preservation::Client.configure(url: Settings.preservation_catalog.url, token: Settings.preservation_catalog.token)
       end
+
+      def aws_client
+        Aws.config.update({
```
jmartin-sul, line 84: i wonder if we could do this configuration in a more centralized way just once, since this is the global config, but i'm blanking on a specific way in the absence of the rails `config/initializers` feature.

jmartin-sul, line 84: another option that wouldn't mess with global config would be to create a one-off `Resource` using the creds from settings, as is done in pres cat: https://github.com/sul-dlss/preservation_catalog/blob/c7b80c3f8ba9931dde1fccd8e846090ec002f94c/app/services/replication/aws_provider.rb#L28-L30

from which you can [get a `bucket` given a bucket name](https://github.com/sul-dlss/preservation_catalog/blob/c7b80c3f8ba9931dde1fccd8e846090ec002f94c/app/services/replication/aws_provider.rb#L18-L20), and then you can get an [`S3Object` from the bucket given the object key](https://github.com/sul-dlss/preservation_catalog/blob/c7b80c3f8ba9931dde1fccd8e846090ec002f94c/app/jobs/replication/delivery_job_base.rb#L16) (which can include slashes so that it's more path like, and i think slashes can be used for something that approximates directory listing, and other directory-ish affordances, but that is a very vague memory).  and then pres cat calls `upload_file` on the `S3Object`, but i actually don't see that in the API documentation?  and it seems like we want to try `write` for streaming anyway.

jmartin-sul: idk if emulating pres cat is the right thing to do.  that code was written a while ago, and i'm not familiar enough with AWS development to know what's idiomatic.  but it seems like parts of that pattern (other than the `upload_file` call) could work well here, and consistency is nice if all else does turn out to be equal.

jmartin-sul (commented):
one other thought besides the ones inline: this or a follow-on PR seem like maybe a good chance to test localstack.  maybe that should be a separate follow-on ticket.

for pres cat, in CI we actually test that the upload method successfully pushes to a real cloud endpoint -- but i think that might be overkill here, esp if localstack ends up being easy.  the mechanism for pushing stuff to the cloud in pres cat has a little more indirection, and downtime feels (to me) a little more tolerable here since failure to send to S3 doesn't block accessioning or put the content at greater risk of getting lost.  plus, i don't think we had the infra integration test suite when we originally wrote those pres cat tests.

lib/dor/text_extraction/file_fetcher.rb
peetucket: Changing the interface/configuration to be more pres like for consistency.

spec/lib/dor/text_extraction/file_fetcher_spec.rb
```diff
@@ -7,26 +7,23 @@
...
-  let(:pres_client) do
-    instance_double(Preservation::Client, objects: objects_client)
-  end
+  let(:base_dir) { "tmp/#{bare_druid}" }
+
+  let(:pres_client) { instance_double(Preservation::Client, objects: objects_client) }
 
   before do
-    FileUtils.mkdir_p(base_dir) unless File.directory?(base_dir)
+    FileUtils.mkdir_p(base_dir)
```
peetucket, line 17: Not needed -- was a comment from the previous PR

jmartin-sul:
part of https://github.com/sul-dlss/common-accessioning/issues/1363

lib/dor/text_extraction/file_fetcher.rb
peetucket: moved to its own AwsProvider class

jmartin-sul (approved):

PR #1357, opened 2024-09-24
remove unncessary directory check

peetucket:

Unnecessary check for directory in the spec.  See https://github.com/sul-dlss/common-accessioning/pull/1349/files#r1772367469


peetucket:
Now part of #1350

Issue #1358, opened 2024-09-24
common-accessioning daemon watches SQS for notifications that content has been STTed, or that an error was encountered in processing


jmartin-sul:
after some discussion and some prototyping in https://github.com/sul-dlss/speech-to-text/pull/9 and https://github.com/sul-dlss/common-accessioning/pull/1367, it seems like we've settled on SQS for communication between SDR and our speech_to_text_generation_service in AWS.  even if SQS is specific to AWS, the general idea of messaging and watching queues should be pretty portable and easy to re-implement on other platforms.

Based on some discussion on 2024-09-25, the plan is to have a daemon in common-accessioning, similar to the `abbyy_watcher` daemon, but instead of monitoring a file system for output, it will watch an SQS queue, and then mark a `speechToTextWF` step accordingly (`stt-create` marked as complete if successful, or failed if an error was encountered during speech to text generation?).

then the workflow can proceed if successful, or the error will surface in the workflow grid if a problem is hit.

this approach allows us to keep the workflow service as an entirely internal application, still only reachable by other SDR services.

~~_**question**_: is it possible/desirable to include the output of successful speech-to-text generation in the queue message or REST request payload?  or is fetching a separate operation from receiving indication of success?~~ we think there's a 256 KB limit on SQS message payload, and https://github.com/sul-dlss/speech-to-text/pull/9 puts the STT output in an S3 bucket, so let's just go forward with retrieval of STT output from S3.

we also considered watching an S3 bucket for incoming files indicating speech-to-text completion, but decided against that since we thought explicit queues or APIs would be more robust and maintainable in the long term.  see https://github.com/sul-dlss/speech-to-text/issues/3#issuecomment-2368453338

see prior edits to this comment if we circle back and want to revisit prior ideas for communicating between SDR and our cloud service.

this is the back into SDR counterpart of https://github.com/sul-dlss/common-accessioning/issues/1356

part of https://github.com/sul-dlss/common-accessioning/issues/1363

jmartin-sul:
notes from https://github.com/sul-dlss/common-accessioning/pull/1367/files#r1775923211
* the speech-to-text job handled by the aws whisper container will attempt to generate text from all of the STT-able files for the druid version being processed, so the queue message to handle may reference many files.
[...]

PR #1367, opened 2024-09-25
Send sqs message for speech to text create

### LINKED ISSUE #1356: `speechToTextWF`: notify speech_to_text_generation_service that there is content to be STTed
discussion and prototyping on https://github.com/sul-dlss/speech-to-text/pull/9 led to us settling on sending messages via SQS, which our cloud speech-to-text container will watch.

this is the push-to-cloud-for-extraction counterpart of https://github.com/sul-dlss/common-accessioning/issues/1358

part of https://github.com/sul-dlss/common-accessioning/issues/1363

peetucket:

Fixes #1356 - send an SQS message to kick off speech to text (after files sent to S3)

New specs

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -59,3 +59,4 @@ aws:
   access_key_id: 'fake-access'
   secret_access_key: 'fake-secret'
   base_s3_bucket: 'sul-speech-to-text-dev'
+  sqs_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url'
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -60,4 +60,5 @@ aws:
   secret_access_key: 'fake-secret'
   speech_to_text:
     base_s3_bucket: 'sul-speech-to-text-dev' # default bucket for storing speech-to-text file
-    sqs_new_job_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url' # queue to send new speech-to-text jobs
+    sqs_todo_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url' # queue to send new speech-to-text jobs
+    sqs_done_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url_done' # queue to receive completed speech-to-text jobs
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -58,5 +58,6 @@ aws:
   region: 'us-west-2'
   access_key_id: 'fake-access'
   secret_access_key: 'fake-secret'
-  base_s3_bucket: 'sul-speech-to-text-dev'
-  sqs_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url'
+  speech_to_text:
+    base_s3_bucket: 'sul-speech-to-text-dev' # default bucket for storing speech-to-text file
+    sqs_new_job_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url' # queue to send new speech-to-text jobs
--- a/lib/robots/dor_repo/speech_to_text/fetch_files.rb
+++ b/lib/robots/dor_repo/speech_to_text/fetch_files.rb
@@ -23,11 +23,7 @@ module Robots
         end
 
         def sttable_filenames
-          @sttable_filenames ||= speech_to_text.filenames_to_stt
-        end
-
-        def speech_to_text
-          @speech_to_text ||= Dor::TextExtraction::SpeechToText.new(cocina_object:, workflow_context: workflow.context)
+          Dor::TextExtraction::SpeechToText.new(cocina_object:, workflow_context: workflow.context).filenames_to_stt
         end
 
         def file_fetcher
--- a/lib/robots/dor_repo/speech_to_text/fetch_files.rb
+++ b/lib/robots/dor_repo/speech_to_text/fetch_files.rb
@@ -4,7 +4,7 @@ module Robots
   module DorRepo
     module SpeechToText
       # Fetch files in need of Speech to Text from Preservation and send to S3
-      class FetchFiles < LyberCore::Robot
+      class FetchFiles < Dor::TextExtraction::Robot
         def initialize
           super('speechToTextWF', 'fetch-files')
         end
@@ -17,10 +17,6 @@ module Robots
         end
 
         private
-
-        def aws_provider
-          @aws_provider ||= Dor::TextExtraction::AwsProvider.new(region: Settings.aws.region, access_key_id: Settings.aws.access_key_id, secret_access_key: Settings.aws.secret_access_key)
-        end
 
         def sttable_filenames
           Dor::TextExtraction::SpeechToText.new(cocina_object:, workflow_context: workflow.context).filenames_to_stt
--- a/lib/robots/dor_repo/speech_to_text/stt_create.rb
+++ b/lib/robots/dor_repo/speech_to_text/stt_create.rb
@@ -30,16 +30,22 @@ module Robots
 
         def send_sqs_message
           message_body = {
-            druid:
+            id: SecureRandom.uuid,
+            druid:,
+            media:
           }.merge(whisper_options).to_json
 
           # Send the message to the SQS queue
           aws_provider.sqs.send_message({
-                                          queue_url: aws_provider.sqs_new_job_queue_url,
+                                          queue_url: aws_provider.sqs_todo_queue_url,
                                           message_body:
                                         })
 
-          logger.info("Sent SQS message for druid #{druid} to queue #{aws_provider.sqs_new_job_queue_url}")
+          logger.info("Sent SQS message for druid #{druid} to queue #{aws_provider.sqs_todo_queue_url}")
+        end
+
+        def media
+          Dor::TextExtraction::SpeechToText.new(cocina_object:, workflow_context: workflow.context).filenames_to_stt
         end
 
         def whisper_options
--- a/lib/robots/dor_repo/speech_to_text/stt_create.rb
+++ b/lib/robots/dor_repo/speech_to_text/stt_create.rb
@@ -11,10 +11,34 @@ module Robots
 
         # available from LyberCore::Robot: druid, bare_druid, workflow_service, object_client, cocina_object, logger
         def perform_work
-          # TODO: do the caption creation by calling out to Whisper
+          # start the caption creation by sending an SQS message to the caption creation service
+          send_sqs_message
 
           # Leave this step running until the Whisper monitoring job marks it as complete
           LyberCore::ReturnState.new(status: :noop, note: 'Initiated SpeechToText.')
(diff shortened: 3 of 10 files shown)


lib/robots/dor_repo/speech_to_text/stt_create.rb
```diff
@@ -11,11 +11,35 @@ def initialize
...
+        private
+
+        def aws_provider
+          @aws_provider ||= Dor::TextExtraction::AwsProvider.new(region: Settings.aws.region, access_key_id: Settings.aws.access_key_id, secret_access_key: Settings.aws.secret_access_key)
+        end
+
+        def send_sqs_message
+          # Define the queue URL and message body
+          queue_url = Settings.aws.sqs_queue_url
+          message_body = {
```
edsu, line 30: The body of the message that speech-to-text is currently looking for is documented here:

https://github.com/sul-dlss/speech-to-text/blob/docker/README.md#the-job-file

edsu (approved):

lib/robots/dor_repo/speech_to_text/stt_create.rb
peetucket: So is the idea that we need to send one message per file that we want transcribed?  if so, then we can iterate over the druid folder in S3, find all of the files, and send multiple messages

though then...how would we know when all of the files in a given druid are done?

jmartin-sul: oh, hmm, that's a good point.  maybe that's an argument for having the speech-to-text service do a directory or a list of files, instead of one file at a time.

alternatively, i can vaguely imagine some ways that the watcher could figure out whether a particular success/failure message for one file, possibly among many for a druid version, is the last one being processed, allowing the step to get marked completed if there were no errors in the batch:
* we could use WFS vars, with e.g. a hash of files to process by the wf for the druid version (hash initialized in this wf step), the processing status updated for each file as it's processed (set upon SQS message receipt by the [watcher daemon we've been discussing today](https://github.com/sul-dlss/common-accessioning/issues/1358)).  and then each time an update is received, the final success/error state of the step could be set?
* we could give common-accessioning its own data store to track that info, but otherwise use similar logic.

i'd probably vote against adding a database to common-accessioning, with no preference between the other two options (and interested in other better options if anyone can think of any).

adding extra load/responsibility to WFS seems not great, but this also seems along the lines of what WFS variables are meant for.  and it doesn't seem like it'd be adding that much load, if we only have a small handful of worker containers processing STT jobs.

keeping the speech-to-text worker container simple also seems desirable, but processing a list of files instead of one file isn't a terrible extra complication (sorry for the back and forth if we do go that route @edsu, i feel like we talked you into one file at a time when we all initially met to look at your work on https://github.com/sul-dlss/speech-to-text/pull/9).

jmartin-sul: given the bare-bones nature of both our argo GUI and workflow-server-rails/lyber-core CLI tools for interacting with robots and workflow state, i'd guess that we shouldn't try to make individual media file extraction failures retriable, and that we should just let the whole batch for the druid version fail if one file fails, then fix the issue and retry the step if need be.  if we run into a bunch of obviously automatically retriable errors, maybe we can revisit and add retry handling for those things where ever is most appropriate?

jmartin-sul: another thought, which could be an optimization: upon successful text generation for a given media file, the watcher process downloads the output for that media file.  then the speech-to-text job queuing code in this class could look to see if there is any output already, and only queue jobs for media files requiring STT that don't already have output.  that saves needless re-running and and thus GPU waste, without having to think about trickier retry capabilities.

jmartin-sul: we'll might need to update https://github.com/sul-dlss/common-accessioning/issues/1358 with whatever comes out of this thread.

jmartin-sul: oh, heh, i may have written all that for nothing: looking back again at the README in https://github.com/sul-dlss/speech-to-text/pull/9/, and at the `download_media` code in that PR for the first time (sort of reviewing these two in parallel, and occasionally flipping over to #1350)... it seems like it already handles a list with multiple files, since the `media` field in the job structure is an array.

i guess the "download all the successful extractions in a partially failed batch" even if it the whole batch doesn't succeed is still a useful thought, and i might've gotten there faster for all the likely useless writing above that 😆

jmartin-sul: also, `SecureRandom.uuid` for the job ID?

config/settings.yml
```diff
@@ -58,4 +58,6 @@ aws:
   region: 'us-west-2'
   access_key_id: 'fake-access'
   secret_access_key: 'fake-secret'
-  base_s3_bucket: 'sul-speech-to-text-dev'
+  speech_to_text:
+    base_s3_bucket: 'sul-speech-to-text-dev' # default bucket for storing speech-to-text file
+    sqs_new_job_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url' # queue to send new speech-to-text jobs
```
jmartin-sul, line 63: what about `sqs_done_queue_url`, to mirror the naming in https://github.com/sul-dlss/speech-to-text/pull/9?

lib/robots/dor_repo/speech_to_text/stt_create.rb
peetucket: Thanks for summarizing.

peetucket:
1. Added a UUID to the message body
2. Added the media files to the message.  

I wonder if instead of reaching out to the cocina object itself to find the list of SSTable files to add to the job, we should instead just look in the AWS bucket folder that should have been created in the previous workflow step and just add any files there.  Could be convinced of either approach, i think they both have pros/cons but probably doesn't matter too much one way or the other.

lib/robots/dor_repo/speech_to_text/fetch_files.rb
```diff
@@ -23,11 +23,7 @@ def aws_provider
         end
 
         def sttable_filenames
-          @sttable_filenames ||= speech_to_text.filenames_to_stt
-        end
-
-        def speech_to_text
-          @speech_to_text ||= Dor::TextExtraction::SpeechToText.new(cocina_object:, workflow_context: workflow.context)
+          Dor::TextExtraction::SpeechToText.new(cocina_object:, workflow_context: workflow.context).filenames_to_stt
         end
```
peetucket, line 27: small refactor ... we don't really need this much indirection since the `filenames_to_stt` is only used once in the robot

lib/robots/dor_repo/speech_to_text/stt_create.rb
```diff
@@ -11,10 +11,51 @@ def initialize
...
+                                          queue_url: aws_provider.sqs_todo_queue_url,
+                                          message_body:
+                                        })
+
+          logger.info("Sent SQS message for druid #{druid} to queue #{aws_provider.sqs_todo_queue_url}")
+        end
+
+        def media
+          Dor::TextExtraction::SpeechToText.new(cocina_object:, workflow_context: workflow.context).filenames_to_stt
+        end
```
peetucket, line 49: this will reach out to DSA again and ask for the list of available files (the same ones we sent to S3 in the previous workflow step)

lib/robots/dor_repo/speech_to_text/stt_create.rb
```diff
@@ -11,10 +11,51 @@ def initialize
...
+        rescue StandardError => e
+          Honeybadger.notify('Problem sending SQS Message to AWS for SpeechToText', context: { druid:, error: e })
+          raise "Error sending SQS message: #{e.message}"
+        end
+
+        private
+
+        def aws_provider
+          @aws_provider ||= Dor::TextExtraction::AwsProvider.new(region: Settings.aws.region, access_key_id: Settings.aws.access_key_id, secret_access_key: Settings.aws.secret_access_key)
+        end
```
peetucket, line 29: i wonder if there is a nicer way to make this `aws_provider` available to any robot that needs it (currently two) without having to reinitialize like this each time (kind of like how we make the workflow_service client available to all robots)

peetucket: Pushed a commit to extract this a new base robot class that we can subclass from.  Not sure how much of an improvement it is other than reducing some duplicated code

config/settings.yml
jmartin-sul: heh, i'd actually meant to say `sqs_todo_queue_url`, but thanks for adding the one i mentioned and doing the rename!

jmartin-sul:
> I wonder if instead of reaching out to the cocina object itself to find the list of SSTable files to add to the job, we should instead just look in the AWS bucket folder that should have been created in the previous workflow step and just add any files there. Could be convinced of either approach, i think they both have pros/cons but probably doesn't matter too much one way or the other.

good question, i could also go either way on that.  maybe post-standup discussion for the devs?

lib/robots/dor_repo/speech_to_text/stt_create.rb
```diff
@@ -4,17 +4,54 @@ module Robots
...
+            media:
+          }.merge(whisper_options).to_json
+
+          # Send the message to the SQS queue
+          aws_provider.sqs.send_message({
+                                          queue_url: aws_provider.sqs_todo_queue_url,
+                                          message_body:
+                                        })
+
+          logger.info("Sent SQS message for druid #{druid} to queue #{aws_provider.sqs_todo_queue_url}")
```
jmartin-sul, line 40: might be nice to log job id also, while we're at it

lib/robots/dor_repo/speech_to_text/stt_create.rb
```diff
@@ -11,10 +11,51 @@ def initialize
...
+        end
+
+        def whisper_options
+          {
+            options: {
+              model: 'large',
+              max_line_count: 80,
+              beam_size: 10
+            }
+          }
```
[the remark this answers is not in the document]
jmartin-sul: i think i'd vote for throwing it in config to start.  as a follow-on touch up.  vague memory of discussion about configuring whisper options (in an argo modal when the workflow is applied?).  if we do that, i could see something where the defaults come from the settings, we merge in overrides from user entry via workflow variable, and then put that combined hash into the options for the message to whisper.

jmartin-sul (approved):

lib/robots/dor_repo/speech_to_text/stt_create.rb
jmartin-sul: https://github.com/sul-dlss/common-accessioning/issues/1369

peetucket: Done in #1370

PR #1373, opened 2024-09-27
adjust job-id params and namespace for sending files to s3

### LINKED ISSUE #1371: SQS Job Id should be druid-version pair
1. When sending files to S3 in the fetch-files step for speech to text, use `#{druid}-v#{version}` instead of just `druid` for the "folder" location of each file (i.e. pre-pending the filename)
2. When sending SQS message, use the same `#{druid}-v#{version}` as the job id instead of a unique UUID (as is current)
3. Do not send the list of files in a "media" block of the SQS message (unnecessary, the job can just look at what's in S3)

peetucket:

Fixes #1371 - job id is now druid-version and so is the place where we put the files in S3 bucket

Spec

### DIFF
--- a/lib/robots/dor_repo/speech_to_text/fetch_files.rb
+++ b/lib/robots/dor_repo/speech_to_text/fetch_files.rb
@@ -12,14 +12,20 @@ module Robots
         # available from LyberCore::Robot: druid, bare_druid, workflow_service, object_client, cocina_object, logger
         def perform_work
           sttable_filenames.each do |filename|
-            raise "Unable to fetch #{filename} for #{druid}" unless file_fetcher.write_file_with_retries(filename:, location: aws_provider.bucket.object(File.join(bare_druid, filename)), max_tries: 3)
+            raise "Unable to fetch #{filename} for #{druid}" unless file_fetcher.write_file_with_retries(filename:, location: aws_provider.bucket.object(File.join(job_id, filename)), max_tries: 3)
           end
         end
 
         private
 
         def sttable_filenames
-          Dor::TextExtraction::SpeechToText.new(cocina_object:, workflow_context: workflow.context).filenames_to_stt
+          Dor::TextExtraction::SpeechToText.new(cocina_object:).filenames_to_stt
+        end
+
+        # this will be the base of the S3 key for the files sent (to namespace them in the bucket)
+        # it is the same as the job_id when we send the SQS message
+        def job_id
+          @job_id ||= Dor::TextExtraction::SpeechToText.new(cocina_object:).job_id
         end
 
         def file_fetcher
--- a/lib/robots/dor_repo/speech_to_text/fetch_files.rb
+++ b/lib/robots/dor_repo/speech_to_text/fetch_files.rb
@@ -11,21 +11,15 @@ module Robots
 
         # available from LyberCore::Robot: druid, bare_druid, workflow_service, object_client, cocina_object, logger
         def perform_work
-          sttable_filenames.each do |filename|
-            raise "Unable to fetch #{filename} for #{druid}" unless file_fetcher.write_file_with_retries(filename:, location: aws_provider.bucket.object(File.join(job_id, filename)), max_tries: 3)
+          speech_to_text.filenames_to_stt.each do |filename|
+            raise "Unable to fetch #{filename} for #{druid}" unless file_fetcher.write_file_with_retries(filename:, location: aws_provider.bucket.object(speech_to_text.s3_location(filename)), max_tries: 3)
           end
         end
 
         private
 
-        def sttable_filenames
-          Dor::TextExtraction::SpeechToText.new(cocina_object:).filenames_to_stt
-        end
-
-        # this will be the base of the S3 key for the files sent (to namespace them in the bucket)
-        # it is the same as the job_id when we send the SQS message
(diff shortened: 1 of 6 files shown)


lib/robots/dor_repo/speech_to_text/fetch_files.rb
```diff
@@ -12,14 +12,20 @@ def initialize
...
-            raise "Unable to fetch #{filename} for #{druid}" unless file_fetcher.write_file_with_retries(filename:, location: aws_provider.bucket.object(File.join(bare_druid, filename)), max_tries: 3)
+            raise "Unable to fetch #{filename} for #{druid}" unless file_fetcher.write_file_with_retries(filename:, location: aws_provider.bucket.object(File.join(job_id, filename)), max_tries: 3)
           end
         end
 
         private
 
         def sttable_filenames
-          Dor::TextExtraction::SpeechToText.new(cocina_object:, workflow_context: workflow.context).filenames_to_stt
+          Dor::TextExtraction::SpeechToText.new(cocina_object:).filenames_to_stt
```
peetucket, line 22: the `filenames_to_stt` method doesn't need workflow_context so don't bother passing it in

lib/robots/dor_repo/speech_to_text/stt_create.rb
```diff
@@ -27,8 +27,7 @@ def perform_work
         def send_sqs_message
           message_body = {
             id: job_id,
-            druid:,
-            media:
+            druid:
```
peetucket, line 30: leaving the druid in the job message (even though it happens to be included in the job_id now) ... maybe useful for logging or other purposes to have it distinct

lib/robots/dor_repo/speech_to_text/stt_create.rb
```diff
@@ -27,8 +27,7 @@ def perform_work
         def send_sqs_message
           message_body = {
             id: job_id,
-            druid:,
-            media:
```
jmartin-sul, line 31: so this got removed in the documentation, but afaict from the code and from testing locally yesterday evening, the `media` field is actually still required in the message body: https://github.com/sul-dlss/speech-to-text/blob/1fd0eb35b721349c75c75be5151dffc8564d4084/speech_to_text.py#L80-L85

i assume the removal from STT documentation was a mistake or a leftover from a discarded development path with the code.  so adding that field back was one of the little touchups i was going to make (to the speech-to-text README) in the small follow-on PR i mentioned this morning at standup.

jmartin-sul: (or, maybe you and ed have discussed and this lines up with a coming change in the speech-to-text python worker?)

spec/robots/dor_repo/speech_to_text/stt_create_spec.rb
```diff
@@ -20,18 +21,17 @@
...
     allow(Aws::SQS::Client).to receive(:new).and_return(aws_client)
     allow(Dor::Services::Client).to receive(:object).and_return(dsa_object_client)
     allow(Dor::TextExtraction::SpeechToText).to receive(:new).and_return(stt)
     allow(LyberCore::WorkflowClientFactory).to receive(:build).and_return(workflow_client)
-    allow(SecureRandom).to receive(:uuid).and_return(job_id)
   end
 
   context 'when the message is sent successfully' do
-    let(:message_body) { { id: job_id, druid:, media: ['file1.mov', 'file2.mp3'], options: { model: 'large', max_line_count: 80, beam_size: 10 } }.to_json }
+    let(:message_body) { { id: job_id, druid:, options: { model: 'large', max_line_count: 80, beam_size: 10 } }.to_json }
```
jmartin-sul, line 34: and then i think this'll need to get reverted if the job json structure isn't actually changing

jmartin-sul (changes requested):
assuming i didn't miss anything about the job JSON on the speech-to-text python worker side intentionally changing, i think this drops a field that's actually needed.  if the same change is in the works on the python side, will approve.

lib/robots/dor_repo/speech_to_text/stt_create.rb
peetucket: Ok, I hadn't further discussed this was @edsu - was operating under the assumption that the whipser container would look for all files that matched the key for the message sent.  

BUT... we could also continue to send the media key here in the message body, and it could basically do this for the container (i.e. look at what is in S3 for that job_id and then send all of those files in the message body)

jmartin-sul: yeah, i could see either approach, and don't think i have a well thought out preference.  seems like an easy adjustment either way... i might bias towards assuming for now that the current worker code is right... i just put up a small PR to that effect in speech-to-text: https://github.com/sul-dlss/speech-to-text/pull/22

unless @edsu has this change to just list bucket contents in the works already?

i think we discussed this at the end of last week, but i can't remember if we settled on an approach or which way we decided if so.

i have to grab lunch after this DLSS all staff that's happening right now, and it'll be 5 eastern time once that's over anyway... maybe we can confer and decide (or remind me of last week's decision 😅) in the morning?  seems like waiting till the morning shouldn't leave anyone seriously blocked this afternoon?

jmartin-sul (approved):

PR #1377, opened 2024-09-30
shell out to try and delete folders instead of FileUtils

peetucket:

We get issues sometimes with an empty folder claiming it cannot be deleted with FileUtils even when it is empty.  This seems to work better (delete happened right away with no retries in my integration test run)

Updated specs.  Deploy to stage and integration tests.

### DIFF
--- a/lib/dor/text_extraction/ocr.rb
+++ b/lib/dor/text_extraction/ocr.rb
@@ -85,7 +85,7 @@ module Dor
 
         files = Dir.glob("#{abbyy_input_path}/*")
         logger.info "Removing ABBYY input directory: #{abbyy_input_path}.  Empty: #{Dir.empty?(abbyy_input_path)}. Num files/folders: #{files.count}: #{files.join(', ')}"
-        FileUtils.rm_r(abbyy_input_path)
+        delete_folder(abbyy_input_path)
       end
 
       # e.g. /abbyy/OUTPUT/ab123cd4567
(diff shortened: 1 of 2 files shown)


lib/dor/text_extraction/ocr.rb
```diff
@@ -85,7 +85,7 @@ def cleanup_input_folder
 
         files = Dir.glob("#{abbyy_input_path}/*")
         logger.info "Removing ABBYY input directory: #{abbyy_input_path}.  Empty: #{Dir.empty?(abbyy_input_path)}. Num files/folders: #{files.count}: #{files.join(', ')}"
-        FileUtils.rm_r(abbyy_input_path)
+        delete_folder(abbyy_input_path)
```
mjgiarlo, line 88: I wonder if `FileUtils.rm_rf` would work here too.

peetucket: It actually is an alias and does the same thing with the same problem: https://ruby-doc.org/stdlib-2.4.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-rm_rf

Adding the `force` param simply ignores the (incorrect) exception and ends up not deleting the empty folder and then proceeding (already tried that).

mjgiarlo (approved):

spec/lib/dor/text_extraction/ocr_spec.rb
```diff
@@ -192,7 +192,7 @@ def build_file(sdr_peserve, filename)
         FileUtils.mkdir_p(ocr.abbyy_output_path)
         allow(Honeybadger).to receive(:notify)
         count = 0
-        allow(FileUtils).to receive(:rm_r) do
+        allow(ocr).to receive(:`) do
```
jmartin-sul, line 195: ha, neat, didn't realize this was the method name for the shell-out-via-backtick syntax, good to know you can intercept the call like this.

peetucket: I didn't either until I asked github copilot and the all knowing AI told me

PR #1378, opened 2024-09-30
Cleanup s3 bucket with speech to text files

### LINKED ISSUE #1374: Implement `SttWorkspaceCleanup` robot in speech-to-text workflow
This robot is analogous to the ocr-workspace-cleanup step.  It will remove the files from S3 at the end of the workflow since captioning is done by then, by finding any files with the `job_id` prefix and removing them from S3.  See #1371 

Note: The clean up should be careful to to list objects with the prefix ending in / because listing keys for /media/bb222cc3333-v2 would include /media/bb222cc3333-v20 ; which is kind of counter intuitive if you start thinking of s3 like a filesystem

peetucket:

Fixes #1374 -- cleanup the s3 bucket that files were uploaded to

Todo: 
- [ ] cleanup the location in s3 where files were captioned to before they were moved back to the workspace

New specs


lib/dor/text_extraction/speech_to_text.rb
```diff
@@ -49,6 +64,11 @@ def job_id
         "#{bare_druid}-v#{cocina_object.version}"
       end
 
+      # return the s3 location for the output files generated by the speech to text service
+      def output_location
+        "#{job_id}/output"
+      end
```
peetucket, line 70: see https://github.com/sul-dlss/speech-to-text/issues/33#issuecomment-2419960981

jmartin-sul (approved):

lib/dor/text_extraction/speech_to_text.rb
```diff
@@ -32,6 +32,21 @@ def required?
...
+      # remove any files in S3 workspace that are no longer needed
+      def cleanup
+        cleanup_s3_folder
+        true
+      end
+
+      # remove all files in the s3 input folder, the output folder is a subfolder, so gets cleaned up too
+      def cleanup_s3_folder
+        # Iterate over the list of filenames to be deleted (which is based on the job_id prefix for this druid)
+        s3_objects = aws_provider.client.list_objects(bucket: aws_provider.bucket_name, prefix: job_id)
```
jmartin-sul, line 44: to deal with the corner case mentioned in the ticket:
> Note: The clean up should be careful to to list objects with the prefix ending in / because listing keys for /media/bb222cc3333-v2 would include /media/bb222cc3333-v20 ; which is kind of counter intuitive if you start thinking of s3 like a filesystem

should prefix actually be `#{job_id}/`?

jmartin-sul: but also, will multiple versions ever be in flight at the same time?  the version is just getting passed around because it makes the workflow service calls easier, right?

peetucket: yes, thanks -- good catch

i fixed in this PR by adding the trailing slash, verified with specs and running on the console to be sure it still worked as expected

PR #1380, opened 2024-10-03
Remove cleanup from end accession

### LINKED ISSUE #1322: Remove the invocation of jobs in accessionWF:end-accession
See https://github.com/sul-dlss/common-accessioning/issues/1289#issuecomment-2197469689  

We have a race condition that is manifested by a potentially long running job invoked by accessionWF:end-accession.  It would be better to not have the job invoked there, but instead by a previous step in accessionWF.  We already have a previous step called reset-workspace that seems to do similar related cleanup work.  This would be a good candidate to move the CleanupJob too (and possibly combine with ResetWorkspaceJob). 

This would be a pre-req: sul-dlss/dor-services-app#5110

Per Captioning work cycle planning meeting 2024-09-25:  This warrants a quick check-in with the wider infra team about whether to proceed, and much (maybe most) of the work will be coordinating testing.  Also, we think that maybe the retries (in DSA job?) will keep the race condition from bothering Andrew with failed workflow steps, but not totally sure about that?

peetucket:

Part of https://github.com/sul-dlss/dor-services-app/issues/5110

Fixes #1322

Remove the call to `cleanup` from end-accession robot.  We will instead do this from the existing reset-workspace robot.  Note that we don't need the call to the `reset` endpoint because all of it's work is already done by cleanup.  The `reset` endpoint is going away in the related DSA and dsa-client PRs.  

We need to coordinate changes in [DSA](https://github.com/sul-dlss/dor-services-app/pull/5183)

Deploy this first before DSA.

Specs
Integration tests

### DIFF
--- a/lib/robots/dor_repo/accession/end_accession.rb
+++ b/lib/robots/dor_repo/accession/end_accession.rb
@@ -13,12 +13,7 @@ module Robots
           next_dissemination_wf = special_dissemination_wf
           workflow_service.create_workflow_by_name(druid, next_dissemination_wf, version: current_version, lane_id:) if next_dissemination_wf.present?
 
-          # Note that this used to be handled by the disseminationWF, which is no longer used.
-          object_client.workspace.cleanup(workflow: 'accessionWF', lane_id:)
-
           start_captioning
-
-          LyberCore::ReturnState.new(status: :noop, note: 'Initiated cleanup API call.')
         end
 
         private
--- a/lib/robots/dor_repo/accession/reset_workspace.rb
+++ b/lib/robots/dor_repo/accession/reset_workspace.rb
@@ -12,8 +12,8 @@ module Robots
         end
 
         def perform_work
-          # Reset workspace is performed async by dor-services-app.
-          object_client.workspace.reset(workflow: 'accessionWF', lane_id:)
+          # Cleanup is performed async by dor-services-app.
+          object_client.workspace.cleanup(workflow: 'accessionWF', lane_id:)
 
           # dor-services-app will update the workflow step, do don't do it here.
           LyberCore::ReturnState.new(status: :noop, note: 'Initiated reset API call.')
--- a/lib/robots/dor_repo/accession/reset_workspace.rb
+++ b/lib/robots/dor_repo/accession/reset_workspace.rb
@@ -4,6 +4,7 @@ module Robots
   module DorRepo
     module Accession
       # This takes link for the object in the /dor/workspace directory and renames it so it has a version number.
+      # It also cleans up the workspace directory.
       # (i.e. /dor/assembly/xw/754/sd/7436/xw754sd7436/ -> /dor/assembly/xw/754/sd/7436/xw754sd7436_v2/)
       class ResetWorkspace < LyberCore::Robot
         def initialize
--- a/lib/robots/dor_repo/dissemination/cleanup.rb
+++ b/lib/robots/dor_repo/dissemination/cleanup.rb
@@ -10,9 +10,9 @@ module Robots
         end
 
         def perform_work
+          object_client.workspace.cleanup(workflow: 'disseminationWF', lane_id:)
+          # Oct 3 2024:  Warn. If we never see these warnings, we can get rid of this robot completely.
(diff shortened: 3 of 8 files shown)


lib/robots/dor_repo/accession/end_accession.rb
```diff
@@ -13,12 +13,7 @@ def perform_work
           next_dissemination_wf = special_dissemination_wf
           workflow_service.create_workflow_by_name(druid, next_dissemination_wf, version: current_version, lane_id:) if next_dissemination_wf.present?
 
-          # Note that this used to be handled by the disseminationWF, which is no longer used.
-          object_client.workspace.cleanup(workflow: 'accessionWF', lane_id:)
-
```
peetucket, line 18: don't make this call from the end-accession.  the work will now be done by reset-workspace robot (well, more specifically, by the job that robot step invokes in DSA once the corresponding DSA changes are done)

lib/robots/dor_repo/accession/end_accession.rb
```diff
@@ -13,12 +13,7 @@ def perform_work
...
           start_captioning
-
-          LyberCore::ReturnState.new(status: :noop, note: 'Initiated cleanup API call.')
```
peetucket, line 21: this step can now complete immediately instead of waiting for the job to complete it -- this prevents the race conditions

lib/robots/dor_repo/dissemination/cleanup.rb
```diff
@@ -3,15 +3,16 @@
...
         def initialize
           super('disseminationWF', 'cleanup')
         end
 
         def perform_work
-          object_client.workspace.cleanup(workflow: 'disseminationWF', lane_id:)
+          Honeybadger.notify('[WARNING] DisseminationWF:cleanup robot is deprecated and should not be used.', context: { druid: })
+          # Oct 3 2024: just warn, do not actually do the work.  If we never see these warnings, we can get rid of this robot.
+          # object_client.workspace.cleanup(workflow: 'disseminationWF', lane_id:)
           LyberCore::ReturnState.new(status: :noop, note: 'Initiated cleanup API call.')
```
peetucket, line 16: this was only place i could find that called the "cleanup" on DSA client (which is going away in https://github.com/sul-dlss/dor-services-client/pull/472)

it shouldn't be called anymore anyway - this robot and workflow has been deprecated for a while now -- so let's just warn HB

lib/robots/dor_repo/accession/reset_workspace.rb
```diff
@@ -4,15 +4,16 @@ module Robots
...
       class ResetWorkspace < LyberCore::Robot
         def initialize
           super('accessionWF', 'reset-workspace')
         end
 
         def perform_work
-          # Reset workspace is performed async by dor-services-app.
-          object_client.workspace.reset(workflow: 'accessionWF', lane_id:)
+          # Cleanup is performed async by dor-services-app.
+          object_client.workspace.cleanup(workflow: 'accessionWF', lane_id:)
```
peetucket, line 16: the reset call is gone -- cleanup does all the work (reset was duplicating what cleanup was doing)

lib/robots/dor_repo/dissemination/cleanup.rb
```diff
@@ -3,15 +3,16 @@
...
+      # This step has been merged into the end accession robot. It should no longer be used.
       class Cleanup < LyberCore::Robot
         def initialize
           super('disseminationWF', 'cleanup')
         end
 
         def perform_work
           object_client.workspace.cleanup(workflow: 'disseminationWF', lane_id:)
+          # Oct 3 2024:  Warn. If we never see these warnings, we can get rid of this robot completely.
+          Honeybadger.notify('[WARNING] DisseminationWF:cleanup robot is deprecated and should not be used.', context: { druid: })
```
peetucket, line 15: we do not think this robot is every invoked - let's add an HB alert, if we don't see that alert ever, we can just ditch this whole robot

jmartin-sul (approved):

jmartin-sul (approved):

jmartin-sul (approved):

PR #1388, opened 2024-10-09
Speech to text watcher

jmartin-sul:

part of #1358

- unit tests
- ran the daemon locally at debug log level
  - configured to point to my own test queues
  - commented out the calls to report to DOR event log and workflow service
  - created and worked a test job using [the tools in speech-to-text for creating test jobs](https://github.com/sul-dlss/speech-to-text?tab=readme-ov-file#run)
  - watched the job get picked up by this daemon, saw the message body get parsed properly based on log statements
  - stopped the daemon and uncommented the calls to report to DOR event log and workflow service
  - created another test job
  - watched the daemon log an error because i don't have my local common-accessioning instance configured to talk to a real dor-services-app or workflow-server-rails instance
  - watched the daemon sit there for the default `visibility_timeout` before picking up the same job again, this time with an increased `ApproximateReceiveCount` (will probably do a follow on to limit retries based on that).  job continued to be retried in that manner till i stopped the daemon again.
- ran integration tests against QA (with this branch deployed and running in a `screen` session).
  - i had trouble when specifying a `role_arn` (which i had to do in my laptop env), but was able to attach to the queue when i omitted the `role_arn`.  that turned up another bug with druid format, fixed in the last commit i pushed on.  i then adjusted the queue names a bit, which i think crashed the speech-to-text container in AWS (when terraform tore down the queues for the old names).  after adjusting the queue and bucket names in the `.env` file for the container and restarting it, i was able to successfully run `spec/features/preassembly_speech_to_text_media_spec.rb` from https://github.com/sul-dlss/infrastructure-integration-test/pull/790 without manual intervention (speech-to-text service running as detached docker container in AWS EC2, watcher started manually in `screen` on common-accessioning-qa-a).

examples created by integration tests (notice completed `speechToTextWF`, and `stt-create-success` event):
- https://argo-qa.stanford.edu/view/vz622bs2090
- https://argo-qa.stanford.edu/view/fb528tv2933

### DIFF
--- a/bin/speech_to_text_watcher
+++ b/bin/speech_to_text_watcher
@@ -1,0 +1,22 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# Continually monitor the speech-to-text worker's "done" queue and report job status to SDR
+
+require 'socket'
+
+require_relative '../config/boot'
+
+# Set log level with LOG_LEVEL; share the logger across the stack
+progname = 'speech_to_text_watcher'
+logger = Logger.new(STDOUT, progname:, level: ENV['LOG_LEVEL'] || 'INFO')
+host = Socket.gethostname
+stt_create_done_handler = Dor::TextExtraction::SpeechToTextCreateDoneHandler.new(host:, progname:, logger:)
+
+
+sqs_watcher = Dor::TextExtraction::SqsWatcher.new(role_arn: Settings.aws.speech_to_text.role_arn,
+                                                   queue_url: Settings.aws.speech_to_text.sqs_done_queue_url)
+sqs_watcher.poll do |done_msg|
+  logger.debug("sqs_watcher.poll block received done_msg: #{done_msg}")
+  stt_create_done_handler.process_done_message(done_msg)
+end
--- a/lib/dor/text_extraction/abbyy/file_watcher.rb
+++ b/lib/dor/text_extraction/abbyy/file_watcher.rb
@@ -22,7 +22,7 @@ module Dor
         # rubocop:disable Metrics/AbcSize
         def initialize(
           logger: Logger.new($stdout),
-          workflow_updater: Dor::TextExtraction::WorkflowUpdater,
+          workflow_updater_class: Dor::TextExtraction::WorkflowUpdater,
           listener_options: {}
         )
           # Set up the ABBYY directories
@@ -33,7 +33,7 @@ module Dor
 
           # Set up the services and listener
           @logger = logger
-          @workflow_updater = workflow_updater.new(logger:)
+          @workflow_updater = workflow_updater_class.new(logger:)
           @listener_options = default_listener_options.merge(listener_options)
           Dor::Services::Client.configure(logger:, url: Settings.dor_services.url, token: Settings.dor_services.token)
         end
--- a/lib/dor/text_extraction/abbyy/file_watcher.rb
+++ b/lib/dor/text_extraction/abbyy/file_watcher.rb
@@ -62,7 +62,7 @@ module Dor
         # Notify SDR that the OCR workflow step completed successfully
         def process_success(results)
           logger.info "Found successful OCR results for druid:#{results.druid} at #{results.result_path}"
-          workflow_updater.mark_ocr_completed("druid:#{results.druid}")
+          workflow_updater.mark_ocr_create_completed("druid:#{results.druid}")
           create_event(type: 'ocr_success', results:)
         end
 
@@ -71,7 +71,7 @@ module Dor
           logger.info "Found failed OCR results for druid:#{results.druid} at #{results.result_path}: #{results.failure_messages.join('; ')}"
           context = { druid: "druid:#{results.druid}", result_path: results.result_path, failure_messages: results.failure_messages }
           Honeybadger.notify('Found failed OCR results', context:)
-          workflow_updater.mark_ocr_errored("druid:#{results.druid}", error_msg: results.failure_messages.join("\n"))
+          workflow_updater.mark_ocr_create_errored("druid:#{results.druid}", error_msg: results.failure_messages.join("\n"))
           create_event(type: 'ocr_errored', results:)
         end
 
--- a/lib/dor/text_extraction/aws_provider.rb
+++ b/lib/dor/text_extraction/aws_provider.rb
@@ -5,14 +5,12 @@ module Dor
     # The Application's configured interface to AWS.
     class AwsProvider
       delegate :client, to: :resource
-      attr_reader :aws_client_args
 
-      def initialize(region:, access_key_id:, secret_access_key:)
-        @aws_client_args = {
-          region:,
-          access_key_id:,
-          secret_access_key:
-        }
+      def initialize(region:, access_key_id:, secret_access_key:, role_arn: nil)
+        @region = region
+        @access_key_id = access_key_id
+        @secret_access_key = secret_access_key
+        @role_arn = role_arn
       end
 
       # @return [::Aws::S3::Bucket]
@@ -37,12 +35,45 @@ module Dor
 
       # @return [::Aws::SQS::Client]
       def sqs
-        @sqs ||= ::Aws::SQS::Client.new(aws_client_args)
+        ::Aws::SQS::Client.new(aws_client_args)
       end
 
       # @return [::Aws::S3::Resource]
       def resource
         ::Aws::S3::Resource.new(aws_client_args)
+      end
+
+      private
+
+      attr_reader :region, :access_key_id, :secret_access_key, :role_arn
+
+      # @return [Hash<Symbol,Object>] a Hash with connection and credential info, for use instantiating AWS client objects (e.g. SQS, S3)
+      # @note We don't memoize the client args, because if they contain an Aws::AssumeRoleCredentials instance, that instance's initial token
+      #   can go stale. Making a new credentials hash is not resource intensive, so we just make a fresh one each time.
+      def aws_client_args
+        aws_client_args = {
+          region:, access_key_id:, secret_access_key:
+        }
+
+        # role_arn handling deals with e.g.
+        # 'Aws::SQS::Errors::AccessDenied: User: arn:aws:iam::1234567890123:user/uname is not authorized
+        # to perform: sqs:receivemessage on resource: arn:aws:sqs:us-west-2:1234567890123:sul-speech-to-text-done
+        # because no resource-based policy allows the sqs:receivemessage action' when polling for SQS messages
+        return aws_client_args unless role_arn.present?
+
+        # https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SQS/Client.html
+        aws_client_args[:credentials] = assume_role_credentials(region:, access_key_id:, secret_access_key:, role_arn:)
+
+        aws_client_args
+      end
+
+      def assume_role_credentials(region:, access_key_id:, secret_access_key:, role_arn:)
+        # https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/setup-config.html#aws-ruby-sdk-credentials-access-token
+        Aws::AssumeRoleCredentials.new(
+          client: Aws::STS::Client.new(region:, access_key_id:, secret_access_key:),
+          role_arn:,
+          role_session_name: 'common-accessioning-session'
+        )
       end
     end
   end
--- a/lib/dor/text_extraction/dor_event_logger.rb
+++ b/lib/dor/text_extraction/dor_event_logger.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Dor
+  module TextExtraction
+    # Send an entry to the dor-services-app event log
+    class DorEventLogger
+      # @param [Logger] an instance of a Ruby Logger (for debug etc logging)
+      def initialize(logger:)
+        Dor::Services::Client.configure(logger:, url: Settings.dor_services.url, token: Settings.dor_services.token)
+      end
+
+      # Publish to the SDR event service with processing information
+      def create_event(druid:, type:, data:)
+        Dor::Services::Client.object(druid).events.create(type:, data:)
+      end
+    end
+  end
+end
--- a/lib/dor/text_extraction/sqs_watcher.rb
+++ b/lib/dor/text_extraction/sqs_watcher.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+module Dor
+  module TextExtraction
+    # Watch an SQS queue for messages by long polling, allowing the consumer
+    # of this class to specify the queue name and the processing logic.
+    class SqsWatcher
+      attr_reader :logger, :queue_url, :role_arn, :visibility_timeout
+
+      def initialize(queue_url:, role_arn: nil, logger: Logger.new($stdout), visibility_timeout: nil)
+        @queue_url = queue_url
+        @role_arn = role_arn
+        @logger = logger
+        @visibility_timeout = visibility_timeout || default_visibility_timeout
+      end
+
+      def poll
+        logger.info("Starting indefinite long polling of #{queue_url}")
+        poller.poll(visibility_timeout:) do |sqs_msg|
+          yield sqs_msg
+        rescue StandardError => e
+          # unexpected error occurred while processing message.
+          # log it, and skip delete so it can be re-processed later.
+          context = { sqs_msg:, error: e }
+          Honeybadger.notify('Error processing SQS message, skipping deletion to allow requeue', context:)
+          logger.error("Error processing SQS message, skipping deletion to allow requeue. Context: #{context}")
+
+          # without this throw, the message would be deleted upon completion of the block, preventing retry
+          throw :skip_delete # TODO: use msg[:attributes]['ApproximateReceiveCount'] to stop retrying above a certain threshold?
+        end
+      end
+
+      private
+
+      def poller
+        options = { client: aws_provider.sqs }
+        Aws::SQS::QueuePoller.new(queue_url, options).tap do |poller|
+          logger.info("created QueuePoller: #{poller.inspect}")
+        end
+      end
+
+      def default_visibility_timeout
+        120 # pick a reasonable default time by which the job should complete successfully, or allow the message to be requeued for retry
+      end
+
+      def aws_provider
+        @aws_provider ||=
+          Dor::TextExtraction::AwsProvider.new(region: Settings.aws.region,
+                                               access_key_id: Settings.aws.access_key_id,
+                                               secret_access_key: Settings.aws.secret_access_key,
+                                               role_arn:)
+      end
+    end
+  end
+end
--- a/lib/dor/text_extraction/workflow_updater.rb
+++ b/lib/dor/text_extraction/workflow_updater.rb
@@ -13,12 +13,12 @@ module Dor
       end
 
       # Notify SDR that the OCR workflow step completed successfully
-      def mark_ocr_completed(druid)
+      def mark_ocr_create_completed(druid)
         @client.update_status(druid:, workflow: OCR_WF_NAME, process: OCR_WF_CREATE_NAME, status: 'completed')
       end
 
       # Notify SDR that the OCR workflow step failed
-      def mark_ocr_errored(druid, error_msg:)
+      def mark_ocr_create_errored(druid, error_msg:)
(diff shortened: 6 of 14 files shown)


lib/dor/text_extraction/sqs_watcher.rb
```diff
@@ -0,0 +1,55 @@
...
+          yield sqs_msg
+        rescue StandardError => e
+          # unexpected error occurred while processing message.
+          # log it, and skip delete so it can be re-processed later.
+          context = { sqs_msg:, error: e }
+          Honeybadger.notify('Error processing SQS message, skipping deletion to allow requeue', context:)
+          logger.error("Error processing SQS message, skipping deletion to allow requeue. Context: #{context}")
+
+          # without this throw, the message would be deleted upon completion of the block, preventing retry
+          throw :skip_delete # TODO: use msg[:attributes]['ApproximateReceiveCount'] to stop retrying above a certain threshold?
```
jmartin-sul, line 29: will file a ticket for this or put a card on the board

lib/dor/text_extraction/dor_event_logger.rb
jmartin-sul, line 1: in a follow on commit or PR, i'll centralize the event creation from `FileWatcher` into this class.

lib/dor/text_extraction/abbyy/file_watcher.rb
```diff
@@ -22,7 +22,7 @@ class FileWatcher
         # rubocop:disable Metrics/AbcSize
         def initialize(
           logger: Logger.new($stdout),
-          workflow_updater: Dor::TextExtraction::WorkflowUpdater,
+          workflow_updater_class: Dor::TextExtraction::WorkflowUpdater,
           listener_options: {}
```
peetucket, line 26: Any reason for this change from `workflow_updater` to `workflow_updater_class` here and in the other references in this file?  It doesn't seem to match what was added in `lib/dor/text_extraction/speech_to_text_create_done_handler.rb` below.

bin/speech_to_text_watcher
```diff
@@ -0,0 +1,22 @@
...
+host = Socket.gethostname
+stt_create_done_handler = Dor::TextExtraction::SpeechToTextCreateDoneHandler.new(host:, progname:, logger:)
+
+
+sqs_watcher = Dor::TextExtraction::SqsWatcher.new(role_arn: Settings.aws.speech_to_text.role_arn,
+                                                   queue_url: Settings.aws.speech_to_text.sqs_done_queue_url)
+sqs_watcher.poll do |done_msg|
+  logger.debug("sqs_watcher.poll block received done_msg: #{done_msg}")
+  stt_create_done_handler.process_done_message(done_msg)
+end
```
peetucket, line 22: how does the watcher get started and stay running?  do we need some hooks in the cap deploy like we do for the abbyy_watcher?

lib/dor/text_extraction/sqs_watcher.rb
```diff
@@ -0,0 +1,55 @@
...
+        @role_arn = role_arn
+        @logger = logger
+        @visibility_timeout = visibility_timeout || default_visibility_timeout
+      end
+
+      def poll
+        logger.info("Starting indefinite long polling of #{queue_url}")
+        poller.poll(visibility_timeout:) do |sqs_msg|
+          yield sqs_msg
+        rescue StandardError => e
```
edsu, line 21: It may be necessary to catch session expiration exceptions here and then run assume role to get an updated session token. Something to keep an eye on if things don't appear to get picked up.

lib/dor/text_extraction/aws_provider.rb
```diff
@@ -7,12 +7,21 @@ class AwsProvider
...
         }
+
+        # role_arn handling deals with e.g.
+        # 'Aws::SQS::Errors::AccessDenied: User: arn:aws:iam::1234567890123:user/uname is not authorized
+        # to perform: sqs:receivemessage on resource: arn:aws:sqs:us-west-2:1234567890123:sul-speech-to-text-done
+        # because no resource-based policy allows the sqs:receivemessage action' when polling for SQS messages
+        return unless role_arn.present?
+
+        # https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SQS/Client.html
+        @aws_client_args[:credentials] = assume_role_credentials(region:, access_key_id:, secret_access_key:, role_arn:)
```
edsu, line 24: I left another note about this below, but one thing I noticed in https://github.com/sul-dlss/speech-to-text/issues/28 is that the assume role operation provides *temporary* session credentials. So I needed to either catch the expired token exception, or not cache the client. I chose to do the latter but maybe I should revisit that.

lib/dor/text_extraction/sqs_watcher.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      def initialize(queue_url:, role_arn: nil, logger: Logger.new($stdout), visibility_timeout: nil)
+        @queue_url = queue_url
+        @role_arn = role_arn
+        @logger = logger
+        @visibility_timeout = visibility_timeout || default_visibility_timeout
+      end
+
+      def poll
+        logger.info("Starting indefinite long polling of #{queue_url}")
```
edsu, line 18: I think `infinite` is what you meant here? :-)

lib/dor/text_extraction/dor_event_logger.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
```
edsu, line 1: I guess this was added here to eventually use it in the abbyy file watcher?

edsu (approved):
This looks like an excellent first pass at the queue watcher -- especially the through tests!

lib/dor/text_extraction/sqs_watcher.rb
peetucket: maybe I'll poll for now, maybe I won't ... got a problem with that?

bin/speech_to_text_watcher
jmartin-sul: oh, possibly, i had overlooked the cap deploy hooks.  there is a corresponding [puppet PR for keeping things running via systemd](https://github.com/sul-dlss/puppet/pull/11366), which i might pare back to just stage for now, leaving QA and prod as a separate draft puppet PR to be held till stage gets properly integration tested with this branch deployed.

jmartin-sul: ok, i see the cap config stuff, i think:
* https://github.com/sul-dlss/common-accessioning/blob/74a2910afe57c401f920add0c21543ddbc2405c6/config/deploy.rb#L46-L47
* https://github.com/sul-dlss/common-accessioning/blob/74a2910afe57c401f920add0c21543ddbc2405c6/config/deploy/stage.rb#L13 (and similar for QA and prod)
* https://github.com/sul-dlss/common-accessioning/blob/main/lib/capistrano/tasks/abbyy_watcher_systemd.cap#L4-L11

i had seen and forgotten the first two things, but had totally missed the third of those.

thinking about it... 
* i think adding the capistrano hooks before the puppet PR is merged would likely break deployment (or at least lead to warnings we don't actually need to worry about), because the cap hooks would refer to some systemd stuff that isn't actually set up yet.
* and i think merging the puppet PR that assumes this watcher script is in place before it's merged to `main` would cause unnecessary warnings, at least, since there'd be attempts to restart a non-existent service.

so my inclination is to ticket the cap config stuff, get this PR merged, then get the puppet PR merged for all envs, then do the ticket for the cap config stuff.

or, maybe we could do all the cap stuff except for the `before 'deploy:publishing', 'abbyy_watcher_systemd:restart'` in `deploy.rb`?

still might be nice to get this merged first.  i'd propose:
* i finish addressing other feedback on this PR
* i deploy this branch to stage and start the daemon manually in a screen session and run the integration test to see if things at least get to marking `stt-create` complete
* i ticket the cap hook stuff for this codebase
* we get this PR merged
* we get the puppet changes merged for all envs
* we unblock the ticket for adding cap hooks to this codebase

lib/dor/text_extraction/abbyy/file_watcher.rb
jmartin-sul: ah, yeah, this was me feeling pedantic and scratching that itch.

the difference between this and `speech_to_text_create_done_handler.rb` is that this class constructor takes the class to instantiate when initializing `@workflow_updater`.  but i still left the actual instance var name as `workflow_updater` like before.  it's just the param name that changes to be more accurate (since it's a class to be instantiated, and it felt nicer to me to not refer to the class on which you call `.new` by the same name as the instance variable that comes from that).

iirc nothing actually overrides the default param value, so maybe i should've just removed the param entirely 🤷 (i might do that in a follow-up PR, just to reduce unnecessary complexity, easy enough to turn back to parameterize this if we need to later).

because nothing that uses it actually sends in an alternate to `WorkflowUpdater`, i didn't see the need to parameterize that in `speech_to_text_create_done_handler.rb`, so it just always does `@workflow_updater = Dor::TextExtraction::WorkflowUpdater.new(logger:)`.

lib/dor/text_extraction/aws_provider.rb
jmartin-sul: ah, good to know -- i'm not sure how to avoid caching for the watcher.  i basically instantiate a poller by using `Aws::SQS::QueuePoller.new(queue_url, options)`, where `options = { client: aws_provider.sqs }` (see `sqs_watcher.rb`).  and then that poller just runs its `#poll` method on the block that handles "done" messages.

ah, but also, i guess this constructor is caching the `Aws::AssumeRoleCredentials` instance, returned by `assume_role_credentials`, as well as the underlying `Aws::STS::Client` instance.

so i guess there's at least one thing to do here, maybe two:
1. something along the lines of turning `aws_client_args` from an instance variable to a method, one where the `:credentials` field is filled in with a fresh result from `assume_role_credentials` every time `aws_client_args` is called. (this `AwsProvider` class)
2. figure out what error to catch, if need be, in the poller, and renew the token in a `rescue` block, or something. (the `SqsWatcher` class)

lib/dor/text_extraction/dor_event_logger.rb
jmartin-sul: yeah, i didn't get to that refactoring, sorry, meant to leave a note.  but my intent is to do a follow up and have the abbyy file watcher switch to this, then it'll be a little less churn to replace the direct call to DSA with the rabbit-mediated dor-services-client equivalent (and not doing that refactoring here makes rebasing my WIP on the rabbit thing from a few weeks ago less of a pain).

i'll ticket that just so it doesn't fall though the cracks (and will mark it blocked by this, or just pick it up myself immediately).

lib/dor/text_extraction/sqs_watcher.rb
jmartin-sul: ha, i had actually thought about a message like `"Polling forever on #{queue_url}"`, but then i figured we actually know that it'll run for a while, but that it'll be interrupted by deploys and outages and such 🙂 

maybe it will actually run for a definite amount of time if i deploy it before addressing the token expiration concern 😆

lib/dor/text_extraction/sqs_watcher.rb
jmartin-sul: ah, yeah, hadn't read this far when i responded up there, but that's what i was thinking in part of my response to the token expiration concern you mentioned in `AwsProvider`.

lib/dor/text_extraction/workflow_updater.rb
```diff
@@ -2,25 +2,36 @@
...
-        @step = step
+      def initialize(client: nil, logger: nil)
         @client = client || LyberCore::WorkflowClientFactory.build(logger:)
       end
 
       # Notify SDR that the OCR workflow step completed successfully
-      def mark_ocr_completed(druid)
-        @client.update_status(druid:, workflow:, process: step, status: 'completed')
+      def mark_ocr_create_completed(druid)
+        @client.update_status(druid:, workflow: OCR_WF_NAME, process: OCR_WF_CREATE_NAME, status: 'completed')
```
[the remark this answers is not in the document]
jmartin-sul: hmm, hadn't thought about that, but thinking it through now...

1. based on a quick read of the workflow-server-rails controller code for `update` (pretty cursory, so not sure of my interpretation): it seems like [current status has to be passed in the params to get a `conflict` error if the status has changed since the update call was made](https://github.com/sul-dlss/workflow-server-rails/blob/4f102d441e525ef42b2cb99bac4b01683852625a/app/controllers/steps_controller.rb#L46).  but we're not doing that here, so i think it'll just get marked `completed`?  i guess that could be problematic since that `update` operation also queues the next step, which would already have been queued if status had already been set.
2. but i think that we'd need to rework the workflow-server-rails controller method a bit to totally prevent race conditions.  if we check the status here before trying to update, we're not guaranteed to avoid the race condition, since the check and update are separate REST calls that can't be wrapped in a transaction.  but even though it'd be a much more fleeting window, there's still an opportunity (afaict) for a race condition on the workflow-server-rails side, because the query result against withc the `current-status` param value is checked isn't obtained in the same transaction in which the update happens.

so... my gut feeling is we should just try our best to avoid this possibility by tuning the visibility timeout?  currently the only thing that uses the SQS poller is `speech_to_text_watcher`, with the `SqsWatcher` default of 120 seconds, and the only thing the `speech_to_text_watcher` has the poller do is update a workflow event and send a DSA event, currently with no retries.  so it feels like that should succeed or fail well within 2 min.  but i'd be happy to just turn it up to 10 or something to be safe.

thinking about it a bit more, i think our latent assumption in most of the SDR queues is that a particular job only gets picked up once, and we're not always great about making job invocations idempotent.  but to be fair, sidekiq superfetch (sidekiq in general?) just guarantees that a job will be run at least once, not exactly once.  so there are weird job timeout situations where things can get double run in other parts of SDR (tho i think that's effectively even more of a corner case than the situation here, and we probably wouldn't run into that sort of thing outside of the sidekiq hotswap timeout being hit for a job spanning a deploy, or some sort of weird redis/worker networking issue between when a job is picked up and when it'd be marked done).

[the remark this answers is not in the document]
peetucket: My contribution to this process is to agree that the `update` operation will happily do nothing if you don't pass in the `current-status`.  But also that if the step gets marked as completed before it's actually done, the workflow will proceed, perhaps incorrectly.

lib/dor/text_extraction/aws_provider.rb
jmartin-sul: ok, i turned `aws_client_args` into a private method, and left a note about why the result isn't cached, here's the diff (making `aws_client_args` private broke the old test, but i was never wild about what was there, and i think the updated version is better:

`````diff
diff --git a/lib/dor/text_extraction/aws_provider.rb b/lib/dor/text_extraction/aws_provider.rb
index 4c151f1..9b06129 100644
--- a/lib/dor/text_extraction/aws_provider.rb
+++ b/lib/dor/text_extraction/aws_provider.rb
@@ -5,23 +5,12 @@ module Dor
     # The Application's configured interface to AWS.
     class AwsProvider
       delegate :client, to: :resource
-      attr_reader :aws_client_args
 
       def initialize(region:, access_key_id:, secret_access_key:, role_arn: nil)
-        @aws_client_args = {
-          region:,
-          access_key_id:,
-          secret_access_key:
-        }
-
-        # role_arn handling deals with e.g.
-        # 'Aws::SQS::Errors::AccessDenied: User: arn:aws:iam::1234567890123:user/uname is not authorized
-        # to perform: sqs:receivemessage on resource: arn:aws:sqs:us-west-2:1234567890123:sul-speech-to-text-done
-        # because no resource-based policy allows the sqs:receivemessage action' when polling for SQS messages
-        return unless role_arn.present?
-
-        # https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SQS/Client.html
-        @aws_client_args[:credentials] = assume_role_credentials(region:, access_key_id:, secret_access_key:, role_arn:)
+        @region = region
+        @access_key_id = access_key_id
+        @secret_access_key = secret_access_key
+        @role_arn = role_arn
       end
 
       # @return [::Aws::S3::Bucket]
@@ -46,7 +35,7 @@ module Dor
 
       # @return [::Aws::SQS::Client]
       def sqs
-        @sqs ||= ::Aws::SQS::Client.new(aws_client_args)
+        ::Aws::SQS::Client.new(aws_client_args)
       end
 
       # @return [::Aws::S3::Resource]
@@ -56,6 +45,28 @@ module Dor
 
       private
 
+      attr_reader :region, :access_key_id, :secret_access_key, :role_arn
+
+      # @return [Hash<Symbol,Object>] a Hash with connection and credential info, for use instantiating AWS client objects (e.g. SQS, S3)
+      # @note We don't memoize the client args, because if they contain an Aws::AssumeRoleCredentials instance, that instance's initial token
+      #   can go stale. Making a new credentials hash is not resource intensive, so we just make a fresh one each time.
+      def aws_client_args
+        aws_client_args = {
+          region:, access_key_id:, secret_access_key:
+        }
+
+        # role_arn handling deals with e.g.
+        # 'Aws::SQS::Errors::AccessDenied: User: arn:aws:iam::1234567890123:user/uname is not authorized
+        # to perform: sqs:receivemessage on resource: arn:aws:sqs:us-west-2:1234567890123:sul-speech-to-text-done
+        # because no resource-based policy allows the sqs:receivemessage action' when polling for SQS messages
+        return aws_client_args unless role_arn.present?
+
+        # https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/SQS/Client.html
+        aws_client_args[:credentials] = assume_role_credentials(region:, access_key_id:, secret_access_key:, role_arn:)
+
+        aws_client_args
+      end
+
       def assume_role_credentials(region:, access_key_id:, secret_access_key:, role_arn:)
         # https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/setup-config.html#aws-ruby-sdk-credentials-access-token
         Aws::AssumeRoleCredentials.new(
diff --git a/spec/lib/dor/text_extraction/aws_provider_spec.rb b/spec/lib/dor/text_extraction/aws_provider_spec.rb
index 34b9591..fddeb1b 100644
--- a/spec/lib/dor/text_extraction/aws_provider_spec.rb
+++ b/spec/lib/dor/text_extraction/aws_provider_spec.rb
@@ -49,7 +49,7 @@ describe Dor::TextExtraction::AwsProvider do
     end
 
     it 'configures the credentials with a role_arn value' do
-      expect(provider.aws_client_args[:credentials]).to eq assume_role_credentials
+      expect(provider.sqs.config.credentials).to eq assume_role_credentials
       expect(Aws::AssumeRoleCredentials).to have_received(:new).once
     end
   end
`````

lib/dor/text_extraction/sqs_watcher.rb
jmartin-sul: good news for complexity we don't have to maintain: i started the watcher daemon locally last night (22:50) before ending work for the day, queued and processed a speech to text job this afternoon (13:52) using my local speech-to-text docker container, and saw the daemon attempt to pick up the job 15 hours after it was originally started, with no auth related exception.  the attempt to process the done message by the poller failed, because my local common-accessioning isn't configured to talk to a live DSA or workflow-server-rails, but it was clear the message got picked up with no trouble.  so hopefully the ruby client handles token renewal for us?  bonus: i got to see the message get picked up again a couple minutes later, after the visibility timeout expired, since the exception handling prevented message deletion.

jmartin-sul:
@edsu and @peetucket i still need to file some follow-on tickets based on the discussion above.  but i tested this successfully on QA (using the [WIP integration test](https://github.com/sul-dlss/infrastructure-integration-test/pull/790)).  will file those tickets tomorrow.  i'm not sure if either of you wants to give the last commit a look?  i also made some shared_configs changes, but this watcher code and the speech-to-text container should both be running on up-to-date configs for QA.

peetucket (approved):

lib/dor/text_extraction/dor_event_logger.rb
jmartin-sul: https://github.com/sul-dlss/common-accessioning/issues/1412

bin/speech_to_text_watcher
jmartin-sul: https://github.com/sul-dlss/common-accessioning/issues/1413

lib/dor/text_extraction/dor_event_logger.rb
jmartin-sul: https://github.com/sul-dlss/common-accessioning/issues/1412

Issue #1361, opened 2024-09-24
`speechToTextWF` `update-cocina`: update cocina structural to reference files generated by speech_to_text_generation_service


jmartin-sul:
for now, blocked by the implementation of preceding steps in `speechToTextWF`, but we could probably parallelize this with work on the preceding steps, if we wanted.  we might be able to work off of the `ocrWF` equivalent as an example?

this is the `speechToTextWF` equivalent of [`ocrWF` `update-cocina`](https://github.com/sul-dlss/common-accessioning/blob/main/lib/robots/dor_repo/ocr/update_cocina.rb)

stub code in the `speechToTextWF`: https://github.com/sul-dlss/common-accessioning/blob/main/lib/robots/dor_repo/speech_to_text/update_cocina.rb

part of https://github.com/sul-dlss/common-accessioning/issues/1363

peetucket:
Note: at the moment, this step is implemented using the exact same algorithm used by the ocr update-cocina step, which means it picks up new files in the workspace generated by Whisper, finds resources with base filenames that match, and then adds them to the resource.

We still need to consider:
1. which files get roles (e.g. transcription)
2. if all files should be added or only of a certain mimetype

peetucket:
**Roles:**

".vtt" = "caption"
".txt" = transcription?

jmartin-sul:
captions confirmed to display correctly

jmartin-sul:
we think is done.  if we find bugs in the cocina generation, we can ticket them separately.

PR #1410, opened 2024-11-08
add roles to speech to text files

peetucket:

Part of #1361 - add transcription/caption roles to vtt and txt files

Note that is refactors the `CocinaUpdater` class by making it a super class, and then subclassing the OCR and SpeechToText variants.  This allows us to keep lots of common functionality while overriding those that are different.

~~HOLD for integration testing~~

Spec
Integration test


spec/support/workspace_helper.rb
```diff
@@ -23,22 +23,12 @@
...
+  fixture_file = "001#{File.extname(filename)}"
+  FileUtils.copy(Pathname('spec') / 'fixtures' / 'ocr' / fixture_file, Pathname(workspace_content_dir) / filename)
 end
 
-def copy_fixture_file(fixture_file, filename)
-  FileUtils.copy(Pathname('spec') / 'fixtures' / 'ocr' / fixture_file, Pathname(workspace_content_dir) / filename)
+def create_speech_to_text_file(filename)
+  fixture_file = "file1#{File.extname(filename)}"
+  FileUtils.copy(Pathname('spec') / 'fixtures' / 'speech_to_text' / fixture_file, Pathname(workspace_content_dir) / filename)
 end
```
peetucket, line 34: refactor these helpers methods to make them less copy/pasta

that also accounts for the changes in the spec above when calling these methods

lib/dor/text_extraction/speech_to_text_cocina_updater.rb
```diff
@@ -0,0 +1,20 @@
...
+
+      # set the use attribute based on the mimetype
+      def use(object_file)
+        case object_file.mimetype
+        when 'text/plain'
+          'transcription'
+        when 'text/vtt'
+          'caption'
+        end
+      end
```
edsu, line 17: Do we care about when the mimetype doesn't match?

edsu (approved):
I left a question but this looks good to merge--nice to not have the cooy paste!

lib/dor/text_extraction/speech_to_text_cocina_updater.rb
peetucket: yes, it will be nil in that case, which results in a blank role, which is what we want

if we have other mimetypes later than need specific roles, they can be easily added here

PR #1423, opened 2024-11-19
auto detect language from json file

peetucket:

Fixes https://github.com/sul-dlss/speech-to-text/issues/45 

Grab the json file that is produced by whisper and use it to update cocina with the languageTag attribute.  Set to nil if it can't be parsed.

Note, this PR will also set the languageTag to nil for OCR files.

Example media object with languageTag of `en` as set by Whiser: https://argo-qa.stanford.edu/view/druid:dw502kn2752
Example OCR object with languageTag of nil: https://argo-qa.stanford.edu/view/druid:ry096wm6997


lib/robots/dor_repo/speech_to_text/stage_files.rb
```diff
@@ -37,9 +37,14 @@ def download_file_from_s3(s3_key, local_file_path)
...
-        # only includes .txt and .vtt files
+        # only includes .txt, .vtt and .json files since these are the only ones we need to add to the object or update cocina
         def output_files
-          @output_files ||= aws_provider.client.list_objects(bucket: aws_provider.bucket_name, prefix: s3_output_folder).contents.map(&:key).select { |filename| filename.end_with?('.txt', '.vtt') }
+          @output_files ||= aws_provider.client.list_objects(bucket: aws_provider.bucket_name, prefix: s3_output_folder).contents.map(&:key).select { |filename| filename.end_with?(*stt_file_extensions) }
+        end
+
+        # the file extensions that we are interested in for speech to text output
+        def stt_file_extensions
+          %w[.txt .vtt .json]
```
peetucket, line 47: this ensures we are pulling the json files from S3 into the workspace so they can be parsed later in the cocina updater step

spec/lib/dor/text_extraction/speech_to_text_cocina_updater_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      expect(file.use).to eq 'caption'
+      expect(file.sdrGeneratedText).to be true
+      expect(file.correctedForAccessibility).to be false
+      expect(file.access.view).to be 'world'
+      expect(file.access.download).to be 'world'
+      expect(file.administrative.publish).to be true
+      expect(file.administrative.sdrPreserve).to be true
+      expect(file.administrative.shelve).to be true
+      expect(file.hasMimeType).to eq 'text/vtt'
+      expect(file.languageTag).to eq 'es'
```
peetucket, line 114: proves the language tag is set from our fixture json file

spec/robots/dor_repo/speech_to_text/stage_files_spec.rb
```diff
@@ -40,14 +41,16 @@
...
     end
 
     after { FileUtils.rm_rf(fake_workspace_path) } # cleanup the fake workspace for the next test run
 
-    it 'copies only the vtt and txt files from s3 to local workspace' do
+    # rubocop:disable RSpec/ExampleLength
+    it 'copies only the vtt, json and txt files from s3 to local workspace' do
       # files are not in the local workspace
-      %w[file.txt file.vtt file1.srt].each do |file|
+      %w[file.txt file.vtt file1.json file1.srt].each do |file|
```
peetucket, line 53: proves we are pulling the json file back from s3 now

spec/robots/dor_repo/speech_to_text/stage_files_spec.rb
```diff
@@ -59,6 +62,9 @@
       expect(File.read("#{fake_workspace_path}/content/file1.txt")).to eq(File.read('spec/fixtures/speech_to_text/file1.txt'))
       expect(File.exist?("#{fake_workspace_path}/content/file1.vtt")).to be true
       expect(File.read("#{fake_workspace_path}/content/file1.vtt")).to eq(File.read('spec/fixtures/speech_to_text/file1.vtt'))
+      expect(File.exist?("#{fake_workspace_path}/content/file1.json")).to be true
+      expect(File.read("#{fake_workspace_path}/content/file1.json")).to eq(File.read('spec/fixtures/speech_to_text/file1.json'))
```
peetucket, line 66: proves we are pulling the json file back from s3 now

lib/dor/text_extraction/speech_to_text_cocina_updater.rb
```diff
@@ -6,6 +6,25 @@ module TextExtraction
     class SpeechToTextCocinaUpdater < CocinaUpdater
       private
 
+      # we will skip .json speech to text files, these are only parsed for language to add to cocina
+      def skip_file?(file)
+        file.basename.to_s.start_with?('.') || file.basename.to_s.end_with?('.json')
```
edsu, line 11: Unrelated: should we consider skipping everything but the .vtt and .txt files? Thinking about https://github.com/sul-dlss/speech-to-text/issues/41

Oh nevermind, I see that is done by controlling what output files are pulled back?

lib/dor/text_extraction/speech_to_text_cocina_updater.rb
```diff
@@ -6,6 +6,25 @@ module TextExtraction
...
+      end
+
+      def add_file_to_new_resource(_path)
+        raise 'Dor::TextExtraction::SpeechToTextCocinaUpdater should not be creating new resources for speech to text files'
```
edsu, line 15: I like this safeguard!

lib/dor/text_extraction/speech_to_text_cocina_updater.rb
```diff
@@ -6,6 +6,25 @@ module TextExtraction
...
+      end
+
+      # parse the language from the json file (if found)
+      def language(path)
+        corresponding_json_file = find_workspace_file("#{stem(path)}.json")
+
+        return unless corresponding_json_file
```
edsu, line 22: Shouldn't we always have a JSON file?

edsu (approved):

lib/dor/text_extraction/speech_to_text_cocina_updater.rb
peetucket: Yes, that is a good idea since we could be even more careful here --- instead of adding everything to cocina except certain things, we could add only certain things to cocina.  And similar for OCR.

And yes, this is also guarded against in what we pull back from S3 but no harm in being extra cautious.

lib/dor/text_extraction/speech_to_text_cocina_updater.rb
peetucket: I would hope so.  Perhaps this could be a raise instead so we know about it.

peetucket: Raised an error instead and added a test.

lib/dor/text_extraction/speech_to_text_cocina_updater.rb
peetucket: Made this inclusionary instead of exclusionary (so only specific files are included).  Changed method name to match.

lib/dor/text_extraction/cocina_updater.rb
```diff
@@ -120,6 +120,11 @@ def find_workspace_file(filename)
         path.exist? ? path : nil
       end
 
+      # you can override this in a subclass to only include certain files in the workspace
+      def include_file?(_file)
+        true
```
peetucket, line 125: note: not hit in tests since each subclass currently implements this method

lib/dor/text_extraction/speech_to_text_cocina_updater.rb
edsu: Nice!

PR #1428, opened 2024-11-26
add language tag from cocina to sqs message

### LINKED ISSUE #1427: If an item's cocina specifies the language for a media file, pass that language on to Whisper instead of using Whisper's language detection
The Cocina model has a field for `langaugeTag` where users can specify the language of a file. Currently we only use that field to drive the caption display interface. 

For generating new captions, we want users to be able to specify a language for Whisper to use for transcription. Since a media item could have multiple files in different languages, we need to be able to specify this language on a per-file basis. We propose to use the `languageTag` on each media file to-be-captioned for this purpose.

Example:

[In this QA item](https://argo-qa.stanford.edu/view/druid:vr271pc9432), the language has been set to English on the audio file that would be sent to Whisper. The idea is for Whisper to use that language when generating captions.

Screenshot 2024-11-22 at 12 39 40 PM

Users are already able to edit the language field using the `file_manifest.csv` in Preassembly or the Argo structural metadata editing that uses the same CSV format. So in the near term we do not need to add a UI for language specification.

**Logic**

- Using the existing logic to determine which media files to caption, look for the `languageTag` on those items. Any other files can be ignored.
- Have whisper try to transcribe in that langauge
- Apply the same language value to the VTT/TXT files that come back from Whisper so that it shows up in the caption display UI
- If no language is specified, auto-detect (what we currently do)

**Additional information**

This differs from the OCR approach because in OCR we:

- are able to select multiple languages for detection
- ABBYY (and presumably other OCR tools) detect language in smaller chunks within an item, not on a per-file basis
- set the language at the whole-item level, not per-file

Whisper can only be given a single language.

peetucket:

Fixes #1427 - adds the `languageTag` from cocina to the SQS message for the object so it can be passed through to Whisper.  The languageTag can be set in pre-assembly via the file manifest.

Note that this change must be coordinated with work done for https://github.com/sul-dlss/speech-to-text/issues/51 .  Since the format of the message has changed, the whisper container needs to know how to process and deal with this new format.

Specs

### DIFF
--- a/lib/dor/text_extraction/speech_to_text.rb
+++ b/lib/dor/text_extraction/speech_to_text.rb
@@ -71,6 +71,12 @@ module Dor
         "#{job_id}/output"
       end
 
+      # given a filename, look in the list of files that can be sent for speech to text, examine the cocina structural
+      #  and return the languageTag for the file (or nil if no language is set)
+      def language_tag(filename)
+        stt_files.find { |file| file.filename == filename }&.languageTag
+      end
+
       private
 
       # iterate through cocina structural contains and return all File objects for files that need to be stt'd
--- a/lib/robots/dor_repo/speech_to_text/stt_create.rb
+++ b/lib/robots/dor_repo/speech_to_text/stt_create.rb
@@ -49,7 +49,12 @@ module Robots
         # array of media files in the bucket folder for this job (excluding s3 folders)
         def media
           filenames = aws_provider.client.list_objects(bucket: aws_provider.bucket_name, prefix: job_id).contents.map(&:key).reject { |key| key.end_with?('/') }
-          filenames.map { |filename| { name: filename, options: { language: stt.language_tag(File.basename(filename)) } } }
+          filenames.map do |filename|
+            media_file = { name: filename }
+            language_tag = stt.language_tag(File.basename(filename))
+            media_file[:options] = { language: language_tag } if language_tag
+            media_file
+          end
         end
 
         # pulled from config, could later be overriden by settings in the workflow context
--- a/lib/robots/dor_repo/speech_to_text/stt_create.rb
+++ b/lib/robots/dor_repo/speech_to_text/stt_create.rb
@@ -43,17 +43,22 @@ module Robots
         end
 
         def job_id
-          @job_id ||= Dor::TextExtraction::SpeechToText.new(cocina_object:).job_id
+          stt.job_id
         end
 
         # array of media files in the bucket folder for this job (excluding s3 folders)
         def media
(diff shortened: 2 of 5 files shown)


lib/robots/dor_repo/speech_to_text/stt_create.rb
```diff
@@ -43,18 +43,23 @@ def message_body
...
         end
 
         # pulled from config, could later be overriden by settings in the workflow context
         def whisper_options
           Settings.speech_to_text.whisper.to_h
         end
+
+        def stt
+          @stt ||= Dor::TextExtraction::SpeechToText.new(cocina_object:)
+        end
```
peetucket, line 62: cache this in the object so it can be used in our updated `media` method without re-instantiating it.

spec/lib/dor/text_extraction/speech_to_text_spec.rb
```diff
@@ -45,7 +45,7 @@
 
       context 'when the object has no files that can be STTed' do
         let(:first_fileset_structural) { instance_double(Cocina::Models::FileSetStructural, contains: [text_file]) }
-        let(:second_fileset_structural) { instance_double(Cocina::Models::FileSetStructural, contains: [text_file, text_file]) }
+        let(:second_fileset_structural) { instance_double(Cocina::Models::FileSetStructural, contains: [text_file, text_file2]) }
```
peetucket, line 48: small typo from a previous PR, turned out to have no impact on the test, but just fixing because

lib/dor/text_extraction/speech_to_text.rb
```diff
@@ -71,6 +71,11 @@ def output_location
         "#{job_id}/output"
       end
 
+      # look in resource structural metadata for a given filename and return it's cocina language (or nil if no language is set)
+      def language_tag(filename)
+        stt_files.find { |file| file.filename == filename }&.languageTag
```
peetucket, line 76: question ... do we want the default value to pass to whisper to be nil or `en` ?  this makes it nil, but we could make it `en`

edsu: Preferably there would be no language parameter passed to the service if it is not known.

edsu (approved):

lib/dor/text_extraction/speech_to_text.rb
peetucket: Let me see if I can make that adjustment.

peetucket: changed - so ready for testing when the new container is ready

edsu:
I ran this PR with https://github.com/sul-dlss/speech-to-text/pull/54 deployed and saw that the new job resulted in success. So I think `[HOLD]` can be removed and this can be merged.

PR #1433, opened 2024-12-06
allow bin/console script environment to be set like rails console

peetucket:

So we can start the bin/console script the same way we start a rails console to set the environment (while still supporting the old ways)

Localhost

### DIFF
--- a/bin/console
+++ b/bin/console
@@ -1,9 +1,24 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+require 'optparse'
+
 # To start an pry session with gems and configuration loaded:
-# [robot_root] $ ./bin/console development
-ENV['ROBOT_ENVIRONMENT'] = ARGV.shift unless ARGV.first.nil?
+# [robot_root] $ ./bin/console production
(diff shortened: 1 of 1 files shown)


bin/console
```diff
@@ -1,9 +1,24 @@
...
+
+  OptionParser.new do |opts|
+    opts.on("-e", "--environment ENV", "Environment") do |env|
+      ENV['ROBOT_ENVIRONMENT'] = env
+    end
+  end.parse!
+
+  ENV['ROBOT_ENVIRONMENT'] ||= 'development'
+end
+puts "Loading #{ENV['ROBOT_ENVIRONMENT']} environment..."
```
peetucket, line 22: You can use any way you want, but if you use more than one way, they will picked up in this order of preference:

1. setting the env variable directly (e.g. ROBOT_ENVIRONMENT=production ./bin/console)
2. setting the env variable by putting the environment at the end (e.g. ./bin/console production)
3. setting the env variable by using the rails like -e switch (e.g. ./bin/console -e production)

jmartin-sul (approved):

PR #1439, opened 2024-12-10
detect audio file tracks

### LINKED ISSUE #1436: add pre-processing step that omits files with only silence from being sent to whisper for speech to text extraction
## background

There are two ways this can happen:
* A video file has no audio track
* An audio file, or the audio track in a video file, contains nothing but silence, or minimal constant buzzing, etc

Silent example from @andrewjbtw (this one has no audio track): https://purl.stanford.edu/wz891rq8670 (“A Soviet Ant Takes A Walk”)

## overall goal

The check should first make sure that an audio track is present, and if it isn't, it should mark the video as silent, and not run it through Whisper for text speech-to-text extraction. If the SDR item only has one media file then the workflow will need to be entirely skipped (this is the case with ocrWF).

The check should then examine audio tracks to see if they are entirely silent by the above definition.  No videos that have only silent audio tracks should be run through speech-to-text extraction.

## example approaches in Ruby that requires ffmpeg and no external services

Here are code examples from @dnoneill, using ffmpeg called from Ruby, that will look for whether an audio track is present, and if it is, checks whether there's anything more than background noise, (e.g. a TV buzzing):
* get metadata and audio streams from a video file using ffmpeg: https://github.com/dnoneill/avpd/blob/a80193523558f9dcdc576ad6b9b3a76669ad2d43/app/helpers/av_helper.rb#L6-L10
* using ffmpeg, get data about silent stretches from from a hash with an `audiostreams` key, like the above helper returns ☝️: https://github.com/dnoneill/avpd/blob/a80193523558f9dcdc576ad6b9b3a76669ad2d43/app/helpers/av_helper.rb#L67-L96
  * this line specifically checks to make sure that there is an audio stream, and could be broken out into its own separate check, which we'd run first, to see if there is an audio stream on which to detect silence: https://github.com/dnoneill/avpd/blob/a80193523558f9dcdc576ad6b9b3a76669ad2d43/app/helpers/av_helper.rb#L68

## possible approach for determining lack of audio track using technical-metadata-service instead of ffmpeg

Another way to determine whether a file has an audio track in the first place would be to query our technical-metadata-service.
* This is the techMD controller method to hit: https://github.com/sul-dlss/technical-metadata-service/blob/2a663db6cbfce3153ae5dc55582a3864210d7704/app/controllers/technical_metadata_controller.rb#L16-L20
  * example of argo using that to render the techMD section: https://github.com/sul-dlss/argo/blob/e9dd4922d66e46d86f479286c995b0f988f1ea3c/app/services/techmd_service.rb#L7-L16

If using techMD, you can determine whether a file has audio by looking for an `audio_count` field with a positive value in the `av_metadata` field for the given file.  (you could also look in the `File parts` section for a `file part` that has `audio_metadata`, but looking for `audio_count` in `av_metadata` is probably easier and no less robust)

Example with audio:
argo techmd media file with audio and video tracks

Example without audio (https://argo-qa.stanford.edu/view/druid:qw921qz8979):
Screenshot 2024-12-09 at 1 14 43 PM

## where should this be implemented?

Not sure whether this is part of `Dor::TextExtraction::FileFetcher` and the `Robots::DorRepo::SpeechToText::FetchFiles` step, or a new step, or what.

We're treating this as a different pre-processing step from avoiding running Whisper on long stretches of silence on audio that contains a mixture of speech and long stretches of silence (see #1435 for that more complex pre-processing step -- hopefully this one is significantly easier to implement).

peetucket:

Fixes #1436   Alter the existing `filenames_to_stt` method to reject files that do not meet specific criteria.  This will automatically filter them out before they get sent to whisper (the same as if they were not a valid mimetype for example).

- detect if an audio track exist in a file and is not mostly silent before allowing it to be sent for speech to text
- only files with audio tracks will be sent to whisper
- if no files are left after they are rejected for not having audio tracks, then either (1) if user kicked off the process from Argo, the rest of the speechToTextWF will be skipped with no further action taken and nothing will be sent to whisper or (2) if the process was kicked off via a pre-assembly job, the end-accession step will go into an error state indicating that speech to text is not possible for this object.

HOLD: this work is dependent on this being done first: https://github.com/sul-dlss/technical-metadata-service/pull/572 so that we can use the additional metadata to check it here the same way we check for audio tracks existing

Note: once merged, if you try to start a speechToTextWF and technical metadata is not present as generated by https://github.com/sul-dlss/technical-metadata-service/pull/572, then the workflow will go into an error state.  This means all media objects that may be sent through speechToTextWF will need to have technical metadata regenerated as described in https://github.com/sul-dlss/common-accessioning/issues/1441

Specs
Integration test


lib/robots/dor_repo/speech_to_text/start_stt.rb
```diff
@@ -10,9 +10,6 @@ def initialize
         end
 
         def perform_work
-          # TODO: Note that the `possible?` method is not complete until we further refine the mimetypes available
-          # see https://github.com/sul-dlss/common-accessioning/issues/1346
-          # and lib/dor/text_extraction/speech_to_text.rb#allowed_mimetypes
```
peetucket, line 15: no longer true, remove this TODO

lib/dor/text_extraction/speech_to_text.rb
```diff
@@ -52,8 +52,42 @@ def cleanup_s3_folder
       # return a list of filenames that should be stt'd
       # iterate over all files in cocina_object.structural.contains, looking at mimetypes
       # return a list of filenames that are correct mimetype
+      # then filter out any files that either (1) do not have an audio track or (2) have audio that is mostly silent
       def filenames_to_stt
-        stt_files.map(&:filename)
+        available_files = stt_files.map(&:filename)
+        available_files.select { |filename| has_audio_track?(filename) && has_useful_audio_track?(filename) }
+      end
```
peetucket, line 59: this is what filters the files to send to the speech to text to only those with an audio track that is not mostly silent

lib/dor/text_extraction/speech_to_text.rb
```diff
@@ -52,8 +52,42 @@ def cleanup_s3_folder
...
+      def tech_metadata
+        @tech_metadata ||= begin
+          resp = Faraday.get("#{Settings.tech_md_service.url}/v1/technical-metadata/druid/#{cocina_object.externalIdentifier}") do |req|
+            req.headers['Content-Type'] = 'application/json'
+            req.headers['Authorization'] = "Bearer #{Settings.tech_md_service.token}"
+          end
+          raise "Technical-metadata-service returned #{resp.status} when requesting techmd for #{bare_druid}: #{resp.body}" unless resp.success?
+
+          JSON.parse(resp.body)
+        end
```
peetucket, line 90: this is the code that does that detection by examining technical metadata

note that volume detection level technical metadata is dependent on https://github.com/sul-dlss/technical-metadata-service/pull/572 being merged first and technical metadata updated on any object that needs it

lib/dor/text_extraction/speech_to_text.rb
```diff
@@ -52,10 +52,50 @@ def cleanup_s3_folder
...
+        file_level_tech_metadata(filename)&.dig('av_metadata', 'audio_count')&.positive? || false
+      end
+
+      # using new technical metadata generated in https://github.com/sul-dlss/technical-metadata-service/pull/572
+      # this uses the audio max_volume and mean_volume fields to determine if the audio is mostly silent
+      # rubocop:disable Metrics/CyclomaticComplexity
+      # rubocop:disable Metrics/PerceivedComplexity
+      def has_useful_audio_track?(filename)
+        audio_metadata = file_level_tech_metadata(filename)&.dig('dro_file_parts')&.find { |parts| parts['part_type'] == 'audio' }&.dig('audio_metadata')
+        return false unless audio_metadata && audio_metadata['max_volume'] && audio_metadata['mean_volume']
```
peetucket, line 72: note that this "return false" means that if the audio metadata with max/mean volume is NOT present, it will not send it to speech to text.  this means any existing media objects won't work unless they have technical metadata re-generated.

we could be more inclusive here and instead return true instead of false in that situation ...

edsu (approved):

PR #1449, opened 2025-01-07
rename files when sending to s3 and match correctly when updating cocina

### LINKED ISSUE #1443: Transcription of audio files unexpectedly skipped for test item
From @andrewjbtw on Slack:
> I was looking at the [wind symphony item](https://argo-qa.stanford.edu/view/druid:xx631hs6770) and noticed that Whisper isn’t creating transcriptions for the files labeled “audio 1” through “audio 6". That may be worth ticketing. I’m not aware of other items where Whisper skipped over audio files that were not silent, and I think these files have music in them.

Our silence detection is relatively simple, and should only be skipping A/V files that techMD [has determined to have max and mean volumes below a given threshold](https://github.com/sul-dlss/technical-metadata-service/pull/572).  Even if there's no speech, and we'd just get hallucinations from e.g. instrumental music, we'd still expect music-only files to be processed based on our current logic.

If the issue isn't that common-accessioning is skipping sending these files to the speech-to-text container for processing, but that processing is erroring out on them, we should transfer this issue to https://github.com/sul-dlss/speech-to-text.

peetucket:

Fixes #1443 - deal with scenario of identical filenames with only differing extensions but renaming media files when sending to whisper, and handling the matching of output files when adding to cocina.

Note that this only occurs for captioning, and not OCR...but perhaps could be extended to do this if needed.

~~HOLD for integration testing~~

Go to https://github.com/sul-dlss/infrastructure-integration-test/pull/812 after merging

- [x] specs
- [x] integration test
- [x] HOLD for retest with https://argo-qa.stanford.edu/view/druid:xx631hs6770


lib/dor/text_extraction/cocina_updater.rb
```diff
@@ -240,8 +244,17 @@ def resource_type(path)
         end
       end
 
+      # get the stem of a file path removing the extension (e.g. '/user/path/file1.xml' => 'file1')
       def stem(path)
-        File.basename(path, '.*').to_s
+        File.basename(path, File.extname(path))
+      end
```
peetucket, line 250: same output, just a bit more canonical using all ruby methods

peetucket:
@andrewjbtw to retest with https://argo-qa.stanford.edu/view/druid:xx631hs6770

lib/dor/text_extraction/cocina_updater.rb
```diff
@@ -13,13 +13,14 @@ module TextExtraction
...
-    #    that contains the same filename "stem" and add the file there. e.g.
-    #    "abc123.xml" would be added to the same resource that contains
-    #    "abc123.tiff".
-    #
+    # 4. If the file does not exist in the Cocina already, look for a resource
+    #    that contains the same filename "stem" or the same filename with extension added to the end of the stem
+    #    and add the file there.
+    #    e.g. "abc123.xml" would be added to the same resource that contains "abc123.tiff"
+    #    and "abc123_tiff.xml" would be added to the same resource that contains "abc123.tiff"
+    #    see https://github.com/sul-dlss/common-accessioning/issues/1443 for the logic behind this
```
edsu, line 23: This makes it sound like the renaming will happen for OCR as well as Speech to Text jobs? I thought from conversation in Slack it was only the latter?

edsu (changes requested):
I left a minor suggestion for one of the comments, otherwise looks good!

lib/dor/text_extraction/cocina_updater.rb
peetucket: I can clarify in the code comments.  There are two parts to this:

1. Renaming of files by adding the file name extension into the filename before sending them off to be processed.  This is currently only done only for captioning
2. When updating cocina at the end, the bundling of output files into the appropriate resource, either by matching filenames exactly or by using our new "includes the file extension" rule.  Either way is supported ensuring backwards compatibility.  So non-renamed OCR files will work and renamed caption files will work.  If we were to also rename OCR files before sending them off to ABBYY, this would also work.

This code comment is the cocina updater class, hence why it apples to any scenario.

peetucket: I added a very small tweak to hopefully make it more clear, but i think the code comment explains what the code is actually doing already, and links to the ticket

jmartin-sul (approved):
slight worries about the corner case behavior and readability of `extracted_filename_with_extension`.  but given the unit test coverage here, the integration test coverage in general, and the ability to recover from preservation if we notice issues, i wouldn't hold this up over those concerns.

lib/dor/text_extraction/cocina_updater.rb
```diff
@@ -240,8 +244,17 @@ def resource_type(path)
...
       def stem(path)
-        File.basename(path, '.*').to_s
+        File.basename(path, File.extname(path))
+      end
+
+      # get the full filename of a file path by extracting the extension from the last part of the filename (e.g. '/user/path/file1_mp4.vtt' => 'file1.mp4')
+      def extracted_filename_with_extension(path)
+        basename_with_extension_stem = stem(path)
+        extension_from_filename = basename_with_extension_stem.include?('_') ? basename_with_extension_stem.split('_').last : ''
+        basename_without_extension_stem = basename_with_extension_stem.gsub("_#{extension_from_filename}", '')
```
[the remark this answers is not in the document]
peetucket: Your concerns are noted, and I added a new spec file to verify some behaviors, and made a small tweak as a result which better handles the behavior of a file wit no underscores in it.  I also made some additional clarifications into the code comments.  

I think there is little risk here because (1) this code is only used to add files that come back from SDR or speech to text, so will be limited to those outputs (2) we still prefer an exact filename match (e.g. "some_file_with_underscores.txt" _prefers_ to match with "some_file_with_underscores.tiff" anyway, so a secondary check only happens if it doesn't) and (3) even if the secondary check were run (which would yield "some_file_with.underscores"), that wouldn't throw an error, and would be very unlikely to match something in the existing object.

edsu (approved):

PR #1455, opened 2025-01-22
Submit speech-to-text job to AWS Batch

### LINKED ISSUE #1453: Update stt-create to send message to Batch Queue
In order for speechToTextWF to queue jobs in [AWS Batch](https://aws.amazon.com/batch/) we need to update `stt_create.rb` so that the message is sent to the AWS Batch Queue instead of AWS SQS:

https://github.com/sul-dlss/common-accessioning/blob/6af48875a02918b825c0ede7d89ad94650ee4026/lib/robots/dor_repo/speech_to_text/stt_create.rb#L27-L35

I believe we can use the [aws-sdk-batch](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Batch/Client.html) gem to do the equivalent of:

```python
batch = boto3.client("batch")

result = batch.submit_job(
    jobName=job_id,
    jobQueue=os.environ.get("SPEECH_TO_TEXT_BATCH_JOB_QUEUE"),
    jobDefinition=os.environ.get("SPEECH_TO_TEXT_BATCH_JOB_DEFINITION"),
    parameters={"job": str(json.dumps(job))},
)
```

https://github.com/sul-dlss/speech-to-text/blob/48ca899d7459a8dc24bc3bc9e30e172aacaf79c7/speech_to_text.py#L302-L307

Note, we can drop the SQS TODO queue from the configuration and will need to add configuration for the Batch Job Queue and the Batch Job Definition. The values for these in our AWS dev (QA) environment can be obtained by running `terraform output` in our speech-to-text Terraform configuration:

**batch_job_defintion**: arn:aws:batch:us-west-2:418214828013:job-definition/sul-speech-to-text-qa:1. 
**batch_todo_queue**: arn:aws:batch:us-west-2:418214828013:job-queue/sul-speech-to-text-qa

edsu:

To allow for automatic provisioning of GPU ready EC2 instances we are going to use AWS Batch instead of running a dedicated EC2 instance. This means that work needs to be submitted to a Batch Queue instead of SQS. Since Batch doesn't have any facility for sending notifications on job completion we are still going to keep the "done" SQS queue.

These changes are dependent on Terraform configuration work for AWS Batch in https://github.com/sul-dlss/terraform-aws/pull/1197

Resolves #1453

CI and Integration Tests in QA

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,8 +6,9 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 gem 'activesupport', '~> 7.0'
 gem 'assembly-image', '~> 2.0' # ruby-vips is used by 2.0.0 for improved image processing
 gem 'assembly-objectfile', '~> 2.1'
+gem 'aws-sdk-batch' # used for submitting jobs for speech-to-text workflow
 gem 'aws-sdk-s3' # used for sending files to S3 for the speech-to-text workflow
-gem 'aws-sdk-sqs' # used for sending sqs mssages for the speech-to-text workflow
+gem 'aws-sdk-sqs' # used for receiving sqs messages from the speech-to-text workflow
 gem 'config'
 gem 'dor-services-client', '~> 15.1'
 gem 'dor-workflow-client', '~> 7.0'
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -59,9 +59,11 @@ aws:
   access_key_id: 'fake-access'
   secret_access_key: 'fake-secret'
   speech_to_text:
+    # values below can be found by running `terraform output` in terraform_aws/organizations/{env}/speech_to_text
     role_arn: 'arn:aws:iam::1234567890123:role/DevelopersRole'
-    base_s3_bucket: 'sul-speech-to-text-dev' # default bucket for storing speech-to-text file
-    sqs_todo_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url' # queue to send new speech-to-text jobs
+    base_s3_bucket: 'bucket-name' # default bucket for storing speech-to-text file
+    batch_job_queue: 'arn:aws:batch:us-west-2:1234567890123:job-queue/queue-name' # AWS Batch queue to submit jobs to
+    batch_job_definition: 'arn:aws:batch:us-west-2:1234567890123:job-definition/queue-name' # AWS Batch job definition to use for job
     sqs_done_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url_done' # queue to receive completed speech-to-text jobs
 
 
--- a/lib/dor/text_extraction/aws_provider.rb
+++ b/lib/dor/text_extraction/aws_provider.rb
@@ -24,11 +24,6 @@ module Dor
       end
 
       # @return [String]
-      def sqs_todo_queue_url
-        Settings.aws.speech_to_text.sqs_todo_queue_url
-      end
-
-      # @return [String]
       def sqs_done_queue_url
         Settings.aws.speech_to_text.sqs_done_queue_url
       end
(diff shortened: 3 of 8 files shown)


Gemfile
```diff
@@ -6,8 +6,9 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 gem 'activesupport', '~> 7.0'
 gem 'assembly-image', '~> 2.0' # ruby-vips is used by 2.0.0 for improved image processing
 gem 'assembly-objectfile', '~> 2.1'
+gem 'aws-sdk-batch' # used for submitting jobs for speech-to-text workflow
```
edsu, line 9: we need this to talk to AWS Batch from Ruby

lib/dor/text_extraction/aws_provider.rb
```diff
@@ -23,11 +23,6 @@ def bucket_name
         Settings.aws.speech_to_text.base_s3_bucket
       end
 
-      # @return [String]
-      def sqs_todo_queue_url
-        Settings.aws.speech_to_text.sqs_todo_queue_url
-      end
-
```
edsu, line 30: We no longer need to the todo SQS queue since AWS Batch has its own queue that it manages.

lib/robots/dor_repo/speech_to_text/stt_create.rb
```diff
@@ -11,35 +11,31 @@ def initialize
...
         end
 
-        def message_body
+        def message
           {
             id: job_id,
             druid:,
             media:
-          }.merge(whisper_options).to_json
+          }.merge(whisper_options)
```
edsu, line 38: Sending the batch job requires knowing the id, so it is passed as a hash here insetead of json serialized hash.

config/settings.yml
```diff
@@ -59,9 +59,11 @@ aws:
   access_key_id: 'fake-access'
   secret_access_key: 'fake-secret'
   speech_to_text:
+    # values below can be found by running `terraform output` in terraform_aws/organizations/{env}/speech_to_text
     role_arn: 'arn:aws:iam::1234567890123:role/DevelopersRole'
     base_s3_bucket: 'sul-speech-to-text-dev' # default bucket for storing speech-to-text file
-    sqs_todo_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url' # queue to send new speech-to-text jobs
+    batch_job_queue: 'arn:aws:batch:us-west-2:1234567890123:job-queue/sul-speech-to-text' # AWS Batch queue to submit jobs to
+    batch_job_definition: 'arn:aws:batch:us-west-2:1234567890123:job-definition/sul-speech-to-text:1' # AWS Batch job definition to use for job
     sqs_done_queue_url: 'https://sqs.us-west-2.amazonaws.com/queue_url_done' # queue to receive completed speech-to-text jobs
```
peetucket, line 67: are we sure we want to check in these actual values into the repo here instead of just putting in placeholders and then having the actual values in shared_configs per environment?

peetucket (approved):
Small question about configs, other LGTM

config/settings.yml
edsu: Sure we could put placeholders here. They aren't really secret, but it could be confusing.

edsu: I've updated it with values that look less real, and also updated a test that was expecting the hard coded value.

spec/robots/dor_repo/speech_to_text/stt_create_spec.rb
```diff
@@ -44,24 +46,31 @@
     let(:message_body) { { id: job_id, druid:, media: }.merge(Settings.speech_to_text.whisper.to_h).to_json }
 
     before do
-      allow(aws_client).to receive(:send_message).with({ queue_url: Settings.aws.speech_to_text.sqs_todo_queue_url, message_body: }).and_return(true)
+      allow(aws_batch_client).to receive(:submit_job).with({ props: { message_body: } }).and_return(true)
```
jmartin-sul, line 49: i think i'm slightly confused... where does `props` come from in this structure?  i can't seem to find it in any of the calling code, in either the app or the specs.  (definitely not something i'd hold up the PR over, i suspect i'm missing something, and everything else makes sense to me)

jmartin-sul (approved):
one question, but to resolve what i suspect is me missing something.  looks good to me!

config/settings.yml
jmartin-sul: chiming in to observe that while the `role_arn` values look legit at a glance, they're fake too, they just have the same number of digits as a real `role_arn`.

jmartin-sul: (i would vote for leaving them as-is, but don't feel strongly -- i have a slight preference for very realistic looking example values)

spec/robots/dor_repo/speech_to_text/stt_create_spec.rb
edsu: Good eye, that appears to fluff left over from me trying to get the tests set up right. I'm removing it to avoid further confusion.

Issue #1477, opened 2025-03-31
content-metadata-create should raise error on malformed XML from goobi


andrewjbtw:
**Describe the bug**
Goobi accessioning involves:

- sending content files to SDR
- sending a stubContentMetadata.xm file to SDR that lists all the files to be accessioned with some structural metatdata
- telling SDR to start accessioning, which kicks off assemblyWF

For reasons still to be determined, there are cases where the stubContentMetadata.xml sent from Goobi is incomplete. SDR parses this XML anyway and starts accessioning, which results in an incomplete set of files being accessioned.

Example truncated XML:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<content type="Image">
  <resource>
    <label>Page 1</label>
    <file name="gx425hm8798_0001.tif" />
    <file name="gx425hm8798_0001.pdf" />
    <file name="gx425hm8798_0001.xml" role="transcription" publish="yes" preserve="yes" shelve="yes" />
  </resource>
  <resource>
    <label>Page 2</label>
    <file name="gx425hm8798_0002.tif" />
    <file name="gx425hm8798_0002.pdf" />
    <file name="gx425hm8798_0002.xml" role="transcription" publish="yes" preserve="yes" shelve="yes" />
  </resource>
  <resource>
    <label>Page 3</label>
    <file name="gx425hm8798_0003.tif" />
    <file name="gx425hm8798_0003.pdf" />
    <file name="gx425hm8798_0003.xml" role="transcription" publish="yes" preserve="yes" shelve="yes" />
  </resource>
  <resource>
    <label>Page 4</label>
    <file name="gx425hm8798_0004.tif" />
    <file name="gx425hm8798_0004.pdf" />
    <file name="gx425hm8798_0004.xml" role="transcription" publish="yes" preserve="yes" shelve="yes" />
  </resource>
  <resource>
    <label>Page 5</label>
    <file name="gx425hm8798_0005.tif" />
    <file name="gx425hm8798_0005.pdf" />
    <file name="gx425hm8798_0005.xml" role="transcription" publish="yes" preserve="yes" shelve="yes" />
  </resource>
  <resource>
    <label>Page 6</label>
    <file name="gx425hm8798_0006.tif" />
    <file name="gx425hm8798_0006.pdf" />
    <file name="gx425hm8798_0006.xml" role="transcription" publish="yes" preserve="yes" shelve="yes" />
  </resource>
  <resource>
    <label>Page 7</label>
[...]

PR #1478, opened 2025-04-03
add retries around stub content metadata generation

peetucket:

Follow on for #1477 

We discovered that the stubContentMetadata.xml file seems to be complete if you wait for a bit.

This could be because:
(1) goobi is writing the files at the same time as the make the API call (synchronously) and we have a race condition
(2) our shared mounts are out of date (i.e. goobi writes to a shared file system, but when we try and read the same file from a different VM what mounts that shared file system, it takes some time for the file to appear complete).

If (1) is true, then goobi should fix this on their end and we shouldn't need this code (but it will help in the meantime).
If (2) is true, then this may be the most expedient solution and mimics what we do in other places (like speech to text and OCR work) to get around the shared file system latency issue.

Specs

### DIFF
--- a/lib/robots/dor_repo/assembly/content_metadata_create.rb
+++ b/lib/robots/dor_repo/assembly/content_metadata_create.rb
@@ -11,19 +11,35 @@ module Robots
         end
 
         # Generate the structural metadata for this object from stub content metadata (if present).
+        # rubocop:disable Metrics/MethodLength
         def perform_work
           return LyberCore::ReturnState.new(status: :skipped, note: 'object is not an item') unless assembly_item.item? # not an item, skip
 
           return LyberCore::ReturnState.new(status: :skipped, note: 'No stubContentMetadata to load from the filesystem') unless assembly_item.stub_content_metadata_exists?
 
-          updated = assembly_item.cocina_model.new(structural: assembly_item.convert_stub_content_metadata)
-          assembly_item.object_client.update(params: updated)
+          tries = 0
+          max_tries = 3
+          begin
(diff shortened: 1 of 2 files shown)


lib/robots/dor_repo/assembly/content_metadata_create.rb
```diff
@@ -11,19 +11,35 @@ def initialize
...
-          assembly_item.object_client.update(params: updated)
+          tries = 0
+          max_tries = 3
+          begin
+            updated = assembly_item.cocina_model.new(structural: assembly_item.convert_stub_content_metadata)
+            assembly_item.object_client.update(params: updated)
+          # sometimes the stubContentMetadata.xml is in process of being written and is invalid
+          # this is a workaround that gives it more time to complete
+          # see https://github.com/sul-dlss/common-accessioning/issues/1477
+          rescue RuntimeError => e
```
justinlittman, line 28: Is this the most specific error that you can catch?

justinlittman (approved):
Approve pending consideration of question.

lib/robots/dor_repo/assembly/content_metadata_create.rb
peetucket: Currently we just raise a generic RuntimeError when there is a problem parsing the stub content metadata: https://github.com/sul-dlss/common-accessioning/blob/main/lib/dor/assembly/stub_content_metadata_parser.rb#L47 

We could create a custom exception and then rescue only that, though seems potentially overkill (and we also just raise runtime errors elsewhere in this codebase).  Worst case is we rescue something else we didn't really intend to, retry a couple times and then raise it again anyway and fail.

PR #1481, opened 2025-04-14
temporarily remove any retries

peetucket:

We can add a delay via a config file (see https://docs.goobi.io/workflow-plugins/en/export/goobi-plugin-export-stanford) on the goobi end between when they create the XML file and when they call the start accession API endpoint.  This temporarily removes any retries so we can see if changing the delay on the goobi end makes a difference.  If it does, we could consider remove the retries logic completely (i.e. revert #1478), though it probably doesn't do any harm to leave it in.

### DIFF
--- a/lib/robots/dor_repo/assembly/content_metadata_create.rb
+++ b/lib/robots/dor_repo/assembly/content_metadata_create.rb
@@ -18,7 +18,7 @@ module Robots
           return LyberCore::ReturnState.new(status: :skipped, note: 'No stubContentMetadata to load from the filesystem') unless assembly_item.stub_content_metadata_exists?
 
           tries = 0
-          max_tries = 3
+          max_tries = 0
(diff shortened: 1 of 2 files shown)


lib/robots/dor_repo/assembly/content_metadata_create.rb
```diff
@@ -18,7 +18,7 @@ def perform_work
           return LyberCore::ReturnState.new(status: :skipped, note: 'No stubContentMetadata to load from the filesystem') unless assembly_item.stub_content_metadata_exists?
 
           tries = 0
-          max_tries = 3
+          max_tries = 0
```
peetucket, line 21: yes, i know, this makes it look like it won't do anything at all, but if you examine the slightly wonky logic, it'll do it exactly once and then raise if there an error

PR #1509, opened 2025-07-11
remove dor-workflow-client usage

### LINKED ISSUE #1505: Replace dor-workflow-client with dor-services-client
DSC should now support all of the methods previously provided by DWC.

There is a new lyber-core major version that removes DWC.

Please integration test (either by running an appropriate integration test or via the application).

jmartin-sul:

fixes #1505 (originally opened for testing #242 as discussed in https://github.com/sul-dlss/lyber-core/pull/242#discussion_r2201340393, but was a good start on the actual ticket after the lyber-core major version bump was merged)

- [x] unit tests
- [ ] full integration test suite

### DIFF
--- a/bin/abbyy_catchup
+++ b/bin/abbyy_catchup
@@ -26,13 +26,13 @@ successes, failures = results.partition(&:success?)
 # Manually process the successes and failures; continue on error
 successes.each do |result|
   watcher.process_success(result)
-rescue Dor::WorkflowException, Dor::Services::Client::Error => e
+rescue Dor::Services::Client::Error => e
   logger.error e.message
   next
 end
 failures.each do |result|
   watcher.process_failure(result)
-rescue Dor::WorkflowException, Dor::Services::Client::Error => e
+rescue Dor::Services::Client::Error => e
   logger.error e.message
   next
 end
--- a/lib/dor/text_extraction/workflow_updater.rb
+++ b/lib/dor/text_extraction/workflow_updater.rb
@@ -9,29 +9,34 @@ module Dor
       STT_WF_NAME = 'speechToTextWF'
       STT_WF_CREATE_NAME = 'stt-create'
 
-      # Default is to update the 'ocr-create' step in the 'ocrWF' workflow
-      def initialize(client: nil, logger: nil)
-        @client = client || LyberCore::WorkflowClientFactory.build(logger:)
-      end
-
       # Notify SDR that the OCR workflow step completed successfully
       def mark_ocr_create_completed(druid)
-        @client.update_status(druid:, workflow: OCR_WF_NAME, process: OCR_WF_CREATE_NAME, status: 'completed')
+        workflow(druid:, workflow_name: OCR_WF_NAME, process: OCR_WF_CREATE_NAME).workflow_process.update(status: 'completed')
       end
 
       # Notify SDR that the OCR workflow step failed
       def mark_ocr_create_errored(druid, error_msg:)
-        @client.update_error_status(druid:, workflow: OCR_WF_NAME, process: OCR_WF_CREATE_NAME, error_msg:)
+        workflow(druid:, workflow_name: OCR_WF_NAME, process: OCR_WF_CREATE_NAME).workflow_process.update_error(error_msg:)
       end
 
       # Notify SDR that the speech-to-text workflow step completed successfully
       def mark_stt_create_completed(druid)
-        @client.update_status(druid:, workflow: STT_WF_NAME, process: STT_WF_CREATE_NAME, status: 'completed')
+        workflow(druid:, workflow_name: STT_WF_NAME, process: STT_WF_CREATE_NAME).workflow_process.update(status: 'completed')
       end
 
       # Notify SDR that the speech-to-text workflow step failed
       def mark_stt_create_errored(druid, error_msg:)
-        @client.update_error_status(druid:, workflow: STT_WF_NAME, process: STT_WF_CREATE_NAME, error_msg:)
+        workflow(druid:, workflow_name: STT_WF_NAME, process: STT_WF_CREATE_NAME).workflow_process.update_error(error_msg:)
+      end
+
+      private
+
+      def object_client(druid)
+        Dor::Services::Client.object(druid)
+      end
+
+      def workflow(druid:, workflow_name:, process:)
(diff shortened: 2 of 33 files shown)


spec/robots/dor_repo/release/release_members_spec.rb
```diff
@@ -7,17 +7,22 @@
...
+  let(:workflow_process) { instance_double(Dor::Services::Client::Process, update: true, update_error: true, status: 'queued') }
+  let(:milestones) { instance_double(Dor::Services::Client::Milestones) }
+
+  let(:object_client) { instance_double(Dor::Services::Client::Object, find: cocina_model, members:, release_tags: release_tag_client, workflow: object_workflow, milestones:) }
 
   before do
     allow(Dor::Services::Client).to receive(:object).and_return(object_client)
-    allow(LyberCore::WorkflowClientFactory).to receive(:build).and_return(workflow_client)
-    allow(workflow_client).to receive(:lifecycle).and_return(true, true, true, false)
+    allow(milestones).to receive(:date).and_return(1.day.ago, 1.day.ago, 1.day.ago, nil)
```
jmartin-sul, line 25: i think this is the right way to port this check?

https://github.com/sul-dlss/dor-services-client/blob/1e7581cbfbbb1935ab7136ef7485265760083add/lib/dor/services/client/milestones.rb#L14-L23

```
# Returns the Date for a requested milestone from workflow lifecycle
#
# @param [String] druid object id
# @param [String] milestone_name the name of the milestone being queried for
# @param [Number] version (nil) the version to query for
# @param [Boolean] active_only (false) if true, return only lifecycle steps for versions that have all processes complete
# @return [Time] when the milestone was achieved.  Returns nil if the milestone does not exist
Dor::Services::Client::Milestones#date(milestone_name:, version: nil, active_only: false)
```

lib/robots/dor_repo/release/release_members.rb
```diff
@@ -29,7 +29,7 @@ def members
         end
 
         def published?(member)
-          workflow_service.lifecycle(druid: member.externalIdentifier, milestone_name: 'published', version: member.version).present?
+          object_client.milestones.date(milestone_name: 'published', version: member.version).present?
```
jmartin-sul, line 32: i think this is the right way to port this check?

https://github.com/sul-dlss/dor-services-client/blob/1e7581cbfbbb1935ab7136ef7485265760083add/lib/dor/services/client/milestones.rb#L14-L23

```
# Returns the Date for a requested milestone from workflow lifecycle
#
# @param [String] druid object id
# @param [String] milestone_name the name of the milestone being queried for
# @param [Number] version (nil) the version to query for
# @param [Boolean] active_only (false) if true, return only lifecycle steps for versions that have all processes complete
# @return [Time] when the milestone was achieved.  Returns nil if the milestone does not exist
Dor::Services::Client::Milestones#date(milestone_name:, version: nil, active_only: false)
```

bin/abbyy_catchup
```diff
@@ -26,13 +26,13 @@ successes, failures = results.partition(&:success?)
 # Manually process the successes and failures; continue on error
 successes.each do |result|
   watcher.process_success(result)
-rescue Dor::WorkflowException, Dor::Services::Client::Error => e
+rescue Dor::Services::Client::Error => e
```
jmartin-sul, line 29: does anything replace `Dor::WorkflowException`, or is that just folded into Dor::Services::Client::Error in a post-dor-workflow-client world?

jmartin-sul:
after whittling down the integration test failures with re-runs, i consistently get failures on the text extraction tests: `preassembly_ocr_image_spec.rb`, `preassembly_ocr_document_spec.rb`, `preassembly_speech_to_text_media_spec.rb`.

all seem to run into an error along the lines of `sdr-ingest-transfer : content dir not found for 'testocr-image-only.pdf' when searching '/dor/workspace/jv/387/bd/1730/jv387bd1730/content'`  and then they stay stuck in retrying when i retry the `sdr-ingest-transfer` step.

https://argo-stage.stanford.edu/view/yx374bf4215 and https://argo-stage.stanford.edu/view/rz309bg6326 are a couple where i retried the workflow step and they stayed stuck.

all the rest passed for me, so that's a significant improvement from the friday runs where almost everything was failing.  not sure whether the lyber-core changes or something else accounts for the remaining failures.  will continue to investigate in the morning.

justinlittman (approved):
Approve pending consideration of comment.

lib/dor/text_extraction/workflow_updater.rb
```diff
@@ -9,29 +9,34 @@ class WorkflowUpdater
...
+        workflow(druid:, workflow_name: STT_WF_NAME, process: STT_WF_CREATE_NAME).workflow_process.update_error(error_msg:)
+      end
+
+      private
+
+      def object_client(druid)
+        Dor::Services::Client.object(druid)
+      end
+
+      def workflow(druid:, workflow_name:, process:)
```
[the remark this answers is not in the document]
jmartin-sul: i think the instance in `LyberCore::Robot` is private, so i'm not sure the child classes have access to it (can't remember if access to private methods is inherited, my vague recollection was no).  though i'd considered making it public so that child classes could use it.

but also, i don't think this would be a use case for that, since i believe `WorkflowUpdater` is getting used outside of a robot context, by two text-extraction-related daemons that monitor for completion of extraction by external services (and then update the workflows to push things along):
* https://github.com/sul-dlss/common-accessioning/blob/77c259715e466c94bd0af0bac8dc0fb32d32269e/lib/dor/text_extraction/speech_to_text_create_done_handler.rb#L13
  * used from the daemon script: https://github.com/sul-dlss/common-accessioning/blob/77c259715e466c94bd0af0bac8dc0fb32d32269e/bin/speech_to_text_watcher
* https://github.com/sul-dlss/common-accessioning/blob/77c259715e466c94bd0af0bac8dc0fb32d32269e/lib/dor/text_extraction/abbyy/file_watcher.rb#L25
  * used from the daemon scripts https://github.com/sul-dlss/common-accessioning/blob/77c259715e466c94bd0af0bac8dc0fb32d32269e/bin/abbyy_watcher and https://github.com/sul-dlss/common-accessioning/blob/77c259715e466c94bd0af0bac8dc0fb32d32269e/bin/abbyy_catchup

jmartin-sul:
integration test suite finally passed with the commit @justinlittman threw on

Issue #1495, opened 2025-06-04
Language selections set in Preassembly are not persisting in ABBYY output


estanfor:
Selecting a language for OCR in preassembly-stage doesn't appear to have an effect on the language settings applied. The ABBYY xml.result file shows that the ABBYY default languages (English, French, German, Russian + English, and Spanish) are selected, regardless of which languages were selected during Preassembly:

``` <RecognitionParams AllowUnformattedTextSynthesisMode="true" DocumentAnalysisSpeedQualityMode="Default" UsePullXTextAndRecognizeRest="true" SpecialRecognitionMode="None" AlwaysLookForVerticalTexts="false" AlwaysUseDefaultDAMode="true" ExtractUncoveredXTextAfterAnalysis="false" ProhibitPictureDetection="false" ProhibitTableDetection="false" ProhibitHiddenTextDetection="true" AllowedFontsMode="Default" RecognitionMode="FullPage" VerificationMode="NoVerification" UseBarcodesClassifier="false" LookForBarcodes="false" RecognitionQuality="Fast"><TextType>Normal</TextType><TextType>Typewriter</TextType><Language>English</Language><Language>French</Language><Language>German</Language><Language>Russian+English</Language><Language>Spanish</Language><UserPatternsAndLanguages/></RecognitionParams> ```

**User Impact**
OCR accuracy is lessened because the OCR settings are less specific. There will be a bigger problem if we start doing OCR in languages that are not among the defaults.

**To Reproduce**
1. Start a Preassembly job at https://sul-preassembly-stage.stanford.edu/ using the Image content type. Choose to generate OCR files and select Russian as the OCR language. See screenshot below for settings.
2. On the ABBYY Windows desktop in S:\AbbyyShare\sdr-ocr-stage\RESULTXML, or in /abbyy/RESULTXML on common-accessioning-stage-a, look for your druid's xml.result file. (If possible, turn off the automatic cleanup for xml.result files before doing this. Otherwise, you'll only have a second to open the file before it's cleaned up.)
3. Check the language settings in the xml.result file. Only Russian should appear if the settings applied during preassembly are correctly saved.
4. Check the ALTO output in /abbyy/OUTPUT. Only Russian characters should appear if the settings applied during preassembly are correctly saved.

**Screenshots**

Preassembly settings:
<img width="639" alt="Image" src="https://github.com/user-attachments/assets/4a627f49-7af1-4dc4-8726-2ef26ae88cd7" />

[...]

PR #1515, opened 2025-07-15
Pass in workflow context when creating new version

peetucket:

Fixes https://github.com/sul-dlss/common-accessioning/issues/1495 

1. We need to ensure that when the new version is created with the captioning workflow (be it ocrWF or speechToTextWF), so we are preserving the workflow context from the previous version.  This is because pre-assembly will have set this version context (i.e. the  language selection) in v1, and we need this in the next version so that the ocr/speechToText workflows can use it when running.
2. Add some new specs to verify the abbyy ticket that is created has the languages passed in from the workflow context
3. Fix rubocop deprecation warnings and add new cops

~~HOLD:~~
~~- test again~~
~~- remove TODO debug lines~~
~~- rebase https://github.com/sul-dlss/common-accessioning/pull/1516 and then adjust new fixture files here~~

Specs
Integration test
Manually on QA

### DIFF
--- a/lib/robots/dor_repo/accession/end_accession.rb
+++ b/lib/robots/dor_repo/accession/end_accession.rb
@@ -22,6 +22,14 @@ module Robots
           @current_version ||= object_client.version.current
         end
 
+        # we will pass on any workflow context to the next version when starting OCR or speechToText
+        # so that user selections like language will be available to the OCR/speechToText workflows,
+        # but we will remove the context values that actually starts OCR/speechToText when doing this,
+        # or else we will end up in an infinite loop, where the workflow gets triggered over and over again
+        def workflow_context_for_next_version
+          workflow.context.except('runOCR', 'runSpeechToText')
+        end
+
         # check to see if we need to run OCR or speech-to-text or start workflow if needed
         # they are currently mutually exclusive, you cannot run both OCR and speech-to-text on the same object at the same time
         def start_captioning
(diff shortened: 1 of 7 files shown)


lib/robots/dor_repo/accession/end_accession.rb
```diff
@@ -39,14 +39,12 @@ def start_captioning
...
-            # start OCR workflow
-            workflow_service.create_workflow_by_name(druid, 'ocrWF', version: next_version, lane_id:)
+            workflow_service.create_workflow_by_name(druid, 'ocrWF', version: next_version, lane_id:, context: workflow.context)
           elsif stt.required?
             # user asked for SpeechToText but the object is not speech-to-textable
             raise 'Object cannot have speech-to-text applied' unless stt.possible?
 
-            # start speechToText workflow
-            workflow_service.create_workflow_by_name(druid, 'speechToTextWF', version: next_version, lane_id:)
+            workflow_service.create_workflow_by_name(druid, 'speechToTextWF', version: next_version, lane_id:, context: workflow.context)
```
peetucket, line 47: the actual fix is adding the `context: workflow.context` to the call when creating the new workflows

lib/robots/dor_repo/accession/end_accession.rb
```diff
@@ -22,6 +22,14 @@ def current_version
           @current_version ||= object_client.version.current
         end
 
+        # we will pass on any workflow context to the next version when starting OCR or speechToText
+        # so that user selections like language will be available to the OCR/speechToText workflows,
+        # but we will remove the context values that actually starts OCR/speechToText when doing this,
+        # or else we will end up in an infinite loop, where the workflow gets triggered over and over again
+        def workflow_context_for_next_version
+          workflow.context.except('runOCR', 'runSpeechToText')
```
justinlittman, line 30: Is it dangerous to pass on all context if we start using context more broadly? Should you just pass on the specific context variables that you know you want?

peetucket: could consider this, advantage is that we limit what is passed on, disadvantage is that we'd need to remember to add them here if we add things in pre-assembly which add to the context

i'm a bit ambivalent since this bit of code is only ever triggered if the user specifically asks for OCR or speech to text, so is for a small subset of objects (admittedly the only ones that i am aware of taking advantage of this context feature)

justinlittman (approved):

lib/robots/dor_repo/ocr/xml_ticket_create.rb
```diff
@@ -13,7 +13,7 @@ def initialize
         def perform_work
           workflow_context = workflow.context
           filepaths = Dor::TextExtraction::Ocr.new(cocina_object:, workflow_context:).filenames_to_ocr
-          Dor::TextExtraction::Abbyy::Ticket.new(filepaths:, druid:, ocr_languages: workflow_context[:ocrLanguages]).write_xml
+          Dor::TextExtraction::Abbyy::Ticket.new(filepaths:, druid:, ocr_languages: workflow_context['ocrLanguages']).write_xml
         end
```
peetucket, line 17: another subtle bug, the workflow context keys are strings, not symbols

PR #1521, opened 2025-07-17
try the delete multiple times, and improve logging

peetucket:

Follow on to #1517 - actually turns out the folder is not expected to be empty since there is a single XML file that is generated by abbyy that we do not care about and will not accession and so will be left behind.  BUT: the call to remove the folder doesn't seem to work right away.  This tries it a few times until it actually deletes.

Stage


lib/dor/text_extraction/ocr.rb
```diff
@@ -94,20 +94,22 @@ def cleanup_input_folder
...
-          sleep(3**tries)
-          break if Dir.empty?(abbyy_output_path) || tries > max_tries
+          break if !Dir.exist?(abbyy_output_path) || Dir.empty?(abbyy_output_path) || tries > max_tries
+
+          sleep(2**tries)
         end
 
-        files = Dir.glob("#{abbyy_output_path}/*")
-        logger.info "Removing ABBYY output directory: #{abbyy_output_path}.  Empty: #{Dir.empty?(abbyy_output_path)}. Num files/folders: #{files.count}: #{files.join(', ')}"
         delete_folder(abbyy_output_path)
```
peetucket, line 113: i have futzed with this for hours now, and this actually seems to clear the output folder

i remain mystified as to why it is all needed, but suffice to say that simply trying to force delete the folder, which should delete it and all of it's contents, does not work... nor does delete the files first, and then trying to delete the (what should be) empty folder

I presume it is file system sync issues and it's a real pain

jmartin-sul (approved):

jmartin-sul:
approved but left open since justin's review was requested

justinlittman (approved):

PR #1532, opened 2025-07-25
[HOLD] take into account renamed s3 files when looking in cocina

### LINKED ISSUE #1531: Language tag for files is not picked up and passed to whisper
Due to the way we rename files when putting them into the s3 bucket to avoid filename collisions, we need to be mindful of how we find the language tag in cocina (which uses exact filename matches).  Otherwise we will not locate the correct language tag.

See discussions: https://stanfordlib.slack.com/archives/C06EVDJV45N/p1753396786374039

peetucket:

Fixes #1531 - ensure we pick up language tags from cocina when sending objects to whisper

HOLD for testing

Specs, stage

### DIFF
--- a/lib/dor/text_extraction/speech_to_text.rb
+++ b/lib/dor/text_extraction/speech_to_text.rb
@@ -105,13 +105,26 @@ module Dor
       # This is because whisper currently only uses the basename to determine the output filename, which will cause a problem if we have two
       # input files with the same basename but different extensions (e.g. "file.mp4" and "file.m4a").
       def s3_location(filename)
-        file_extension = File.extname(filename)
-        basename = File.basename(filename, file_extension)
+        File.join(job_id, s3_filename(filename))
+      end
+
+      def s3_filename(filename)
         # For example, "file1.mp4" becomes "file1_mp4.mp4" which will cause whipser to produce "file1_mp4.vtt" as the output,
         # and "file1.m4a" becomes "file1_m4a.m4a" which will cause whisper to produce "file1_m4a.vtt" as the output, ensuring uniqueness for both.
         # The cocina_updater will understand this and update the cocina structural metadata correctly matching the new filename to the original.
-        new_filename = "#{basename}_#{file_extension.delete('.')}#{file_extension}"
-        File.join(job_id, new_filename)
+        file_extension = File.extname(filename) # e.g. .mp4
+        basename = File.basename(filename, file_extension) # e.g. file1
+        "#{basename}_#{file_extension.delete('.')}#{file_extension}" # e.g. file1_mp4.mp4
+      end
+
+      def cocina_filename(filename)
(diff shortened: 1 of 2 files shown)


lib/dor/text_extraction/speech_to_text.rb
```diff
@@ -105,13 +105,26 @@ def tech_metadata
       # This is because whisper currently only uses the basename to determine the output filename, which will cause a problem if we have two
       # input files with the same basename but different extensions (e.g. "file.mp4" and "file.m4a").
       def s3_location(filename)
-        file_extension = File.extname(filename)
-        basename = File.basename(filename, file_extension)
```
peetucket, line 109: refactored this part into it's own method and documented with that method

lib/dor/text_extraction/speech_to_text.rb
```diff
@@ -124,10 +137,11 @@ def output_location
...
       end
 
-      # given a filename, look in the list of files that can be sent for speech to text, examine the cocina structural
+      # given a s3 filename, look in the list of files that can be sent for speech to text, examine the cocina structural
       #  and return the languageTag for the file (or nil if no language is set)
-      def language_tag(filename)
-        stt_files.find { |file| file.filename == filename }&.languageTag
+      #  we need to account for the way we rename files by adding the extension when sending to S3
+      def language_tag(s3_filename)
+        stt_files.find { |file| file.filename == cocina_filename(s3_filename) }&.languageTag
```
peetucket, line 144: use our new method so that when we have an s3 filename, we can find the matching one in cocina

spec/lib/dor/text_extraction/speech_to_text_spec.rb
```diff
@@ -105,28 +105,41 @@
...
+      end
+    end
+
+    context 'when the file is found, the s3 filename is the same as the cocina filename and there is a language tag in cocina' do
+      let(:filename) { 'file1.mp4' }
+      let(:s3_filename) { 'file1.mp4' }
+
+      it 'returns the language tag' do
+        expect(stt.filenames_to_stt).to include(filename)
+        expect(stt.language_tag(s3_filename)).to eq 'es'
```
peetucket, line 142: our new method will also work even if the filenames match exactly between s3 and cocina

lib/dor/text_extraction/speech_to_text.rb
edsu: Ah, so this is why the language tag wasn't being found -- because it was looking it up using the original filename instead of the one for s3?

edsu (approved):

lib/dor/text_extraction/speech_to_text.rb
peetucket: yes, exactly -- and it wasn't erroring, it just wasn't finding the language tag and so was never adding it.

PR #1535, opened 2025-07-29
Create releaseWF for members not the collection.

justinlittman:
releaseWF was being created for the collection, not the members. This was causing widespread mayhem.

Unit

### DIFF
--- a/lib/robots/dor_repo/release/release_members.rb
+++ b/lib/robots/dor_repo/release/release_members.rb
@@ -29,7 +29,8 @@ module Robots
         end
 
         def published?(member)
-          object_client.milestones.date(milestone_name: 'published', version: member.version).present?
+          # This is for the member, not the parent collection.
+          Dor::Services::Client.object(member.externalIdentifier).milestones.date(milestone_name: 'published', version: member.version).present?
         end
 
         # Here's an example of the kinds of tags we're dealing with:
@@ -43,7 +44,8 @@ module Robots
         end
 
         def create_release_workflow(member)
-          object_workflow.create(version: member.version, lane_id:)
+          # This is for the member, not the parent collection.
+          Dor::Services::Client.object(member.externalIdentifier).workflow('releaseWF').create(version: member.version, lane_id:)
         end
       end
     end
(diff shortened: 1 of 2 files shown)


spec/robots/dor_repo/release/release_members_spec.rb
```diff
@@ -16,13 +17,13 @@
...
+  # let(:milestones) { instance_double(Dor::Services::Client::Milestones) }
 
-  let(:object_client) { instance_double(Dor::Services::Client::Object, find: cocina_model, members:, release_tags: release_tag_client, workflow: object_workflow, milestones:) }
+  let(:object_client) { instance_double(Dor::Services::Client::Object, find: cocina_model, members:, release_tags: release_tag_client, workflow: object_workflow) }
 
   before do
-    allow(Dor::Services::Client).to receive(:object).and_return(object_client)
-    allow(milestones).to receive(:date).and_return(1.day.ago, 1.day.ago, 1.day.ago, nil)
+    allow(Dor::Services::Client).to receive(:object).with(druid).and_return(object_client)
+    # allow(milestones).to receive(:date).and_return(1.day.ago, 1.day.ago, 1.day.ago, nil)
```
edsu, line 26: should this be removed?

lib/robots/dor_repo/release/release_members.rb
```diff
@@ -29,7 +29,7 @@ def members
         end
 
         def published?(member)
-          object_client.milestones.date(milestone_name: 'published', version: member.version).present?
+          Dor::Services::Client.object(member.externalIdentifier).milestones.date(milestone_name: 'published', version: member.version).present?
```
edsu, line 32: I don't see what the difference is here. Is it worth a comment--especially since it seems like a pretty important difference?

spec/robots/dor_repo/release/release_members_spec.rb
jcoyne, line 26: Cruft?

edsu (approved):
You probably want someone with more knowledge of what's going on here to review, but I left a few suggestions and am marking approved since it sounds like it's important to get released?

PR #1537, opened 2025-07-29
use object access rights when adding ocr/speech to text files

### LINKED ISSUE #1536: Files created by OCR / text extraction should follow the object rights
**Describe the bug**
When the ocrWF creates new files (OCR text, PDF), those files should be accessioned with the same rights as the object.

Currently, all OCR files are being set to world even if the object rights are restricted. See https://argo.stanford.edu/view/druid:th601pc2412

**User Impact**
Items with restricted rights will become downloadable unintentionally.

**To Reproduce**
1. Go to an item with restricted rights.
2. Run text extraction
3. See that the files become world-downloadable.

**Expected behavior**
The OCR workflow should not set file rights different from the object rights.

**Additional context**
I thought we already had this set up when we did the OCR workcycle but I'm not finding a previous issue.

peetucket:

Fixes #1536 - access rights for new files generated by OCR/speech-to-text should match the default object rights, and not "world/world"

New specs

Todo:
 - [x] run integration tests
 - [x] run other tests on stage

### DIFF
--- a/lib/dor/text_extraction/cocina_updater.rb
+++ b/lib/dor/text_extraction/cocina_updater.rb
@@ -201,10 +201,11 @@ module Dor
         nil
       end
 
+      # you can override this in a subclass to set the access attributes conditionally if needed, otherwise uses object default
       def access
         {
-          view: 'world',
-          download: 'world'
+          view: dro.access.view,
+          download: dro.access.download
         }
       end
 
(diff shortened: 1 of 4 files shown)


peetucket:
Integration tests run and Andrew verified fix in stage

aaron-collier (approved):

